PR #21, opened 2020-05-18
redescribed the protocol instead of starting from how it was invented

colourful-land:
- Replace cryptic descriptions like 'holds an assertion of' with the actual steps.
- Remove the old protocol before Improvement.
- Describe step-by-step, e.g. previously omitted what smart contract checks, only what smart contract is capable of checking.


jot2re:
One minor thing that should be mentioned is that it is very important that the identifier is the output of a cryptographic hash function. If not, then it is possible for a blockchain node to steal the money from Bob by picking a specific identifier and get an attestation to that. However, if the identifier is the output of a hash function he will have to find a preimage of a hash function to be able to carry out that attack.

PR #27, opened 2020-06-25
this version of X-1 converter works with every known ASN.X we had exc…

colourful-land:
Hi @jot2re try it on your computer and please approve and merge this if it works with the old ASN.X. Even if it doesn't work with your modified ASN.X files in another PR https://github.com/AlphaWallet/blockchain-attestation/pull/26


colourful-land:
> @colourful-land could you bring this feedback with issues to the developer of the XSL file? 
> Also please let me know if I should merge given that the ASN resulting from the XSL is not completely valid.

1. We should merge this PR as long as the translator works correctly for the files in the existing master branch (not the newly added ones which breaks the translator).

2. We should merge this PR as long as the translator works correctly in terms of translation, not in terms of X.509 compatibility. So for example "the critical variable in Extension is optional" does not affect this PR, instead, file a separate issue against the ASD files which got it wrong in the first place.

3. I am going to get the developer to jump in this PR but we can't expect the developer of this XSL to understand RFC 5280 or X.509. He is supposed to work on the text level. So point out "the translation should be 𝑎 but it is 𝑏" instead of explaining what 𝑎 or 𝑏 means in the context of X.509. One way to think about it is to imagine the developer being a translation AI. Google Translator for example never really understood what a teapot is or what is the taste of tea.

Once you separate your comments into 𝑎) an issue against ASD, mention RFC5280 when you need; 𝑏) a list of expectation/reality of the translator, without mentioning RFC5280, I'll ask the developer of the XLS to jump in and fix the translator.

jot2re:
I am going to wait with merging then since there are two syntax issue with the the produced ASN modules:
1. In InformationFramework the line `WITH SYNTAX { [&Type]  IDENTIFIED BY &id }` should be `WITH SYNTAX { &Type  IDENTIFIED BY &id }` to be syntactically correct. 
2. Similarly in AuthenticationFramework the line `WITH SYNTAX { [&ExtnType] [CRITICAL &critical] IDENTIFIED BY &id }` should be `WITH SYNTAX { &ExtnType CRITICAL &critical IDENTIFIED BY &id }` to be syntactically correct.

I have not made a new issue about the inconsistencies with RFC 5280 since I already tried to implement these in PR 26. If you can get the developer to have a look at that PR then I will write the issues I have with the XSL in the comments there (so that the description of the issues matches the files in the PR where they are described).

colourful-land:
> I am going to wait with merging then since there are two syntax issue with the the produced ASN modules:
> 
>     1. In InformationFramework the line `WITH SYNTAX { [&Type]  IDENTIFIED BY &id }` should be `WITH SYNTAX { &Type  IDENTIFIED BY &id }` to be syntactically correct.
> 
>     2. Similarly in AuthenticationFramework the line `WITH SYNTAX { [&ExtnType] [CRITICAL &critical] IDENTIFIED BY &id }` should be `WITH SYNTAX { &ExtnType CRITICAL &critical IDENTIFIED BY &id }` to be syntactically correct.
> 

Check if the updated version is correct now! Thanks!

jot2re:
Yes! Thanks. It validates correctly now.

PR #34, opened 2020-11-09
Attestation demo

colourful-land:
Quote from Tore:
----

I have added it as a separate package in the blockchain-attestation repo on a new branch called attestation-demo. 
Furthermore I have updated the Gradle to construct a fat, executable jar that can run the demo completely stand-alone. It is constructed with the command “shadowJar”. 

I have also added a test class called testDemo (in the attestation-demo branch on blockchain-attestations) which executes all the command line commands for a complete flow. 

I have made a copy of all the attestation code from the android repo and pasted that into the blockchain-attestation repo. I have tried to compile a library jar that could then be used in the android repo. However my newly acquired "Gradle skills” does not seem to be at that level yet, since I still get the issue of “package not found” when I try to use the attestation code in such a compiled jar in another project. 
Maybe one of you knows exactly what I am missing to make that work? What I tried was the following:

    configurations {
        externalLibs
    }
    externalLibs files('lib/blockchain-attestation.jar')
    implementation files('lib/blockchain-attestation.jar')

Kind Regards,


jot2re:
I have now gotten the import of the attestation jar constructed in this project in the android alphawallet repo working. So disregard the issue mentioned above. 
Basically running packageSkinny constructs a jar: build/libs/blockchain-attestation.jar. This can then be used in any other project by copying the jar file and adding the line  `implementation files('blockchain-attestation.jar')` to the dependencies in Gradle.

PR #38, opened 2020-11-27
all pieces of ASN.X modules needed for attestation framework except C…

James-Sangalli:
…SR which will be separate


James-Sangalli:
Closing to keep branch recoverable

PR #43, opened 2020-12-08
Normalised coordiantes

jot2re:
The issue Oleg mentioned was because points were not normalised in the java implementation but the JS library normalised point by default. I have made a change to normalise things in the Java implementation as well.


colourful-land (approved):
Thanks

PR #45, opened 2020-12-10
Tests and improvements

jot2re:
Since we are constructing other code based on the java reference and that this will actually be used in the real world I decided to write more tests. In particular negative tests to ensure that it is not possible to subvert the checks we expect to have in place. 
I didn't find any significant issues, only minor corrections and better exception handling


colourful-land (approved):
Whoo a lot of update here thanks!

colourful-land:
It's merged in 736230fd66d247dddd3e684b4072ec6e2972d4f9

PR #49, opened 2020-12-14
Nick playground (draft stage)

nicktaras:
draft stage change for review


playground/index.html
```diff
@@ -0,0 +1,52 @@
...
+    </div>
+  </div>
+
+  <script>
+    var tokens = [];
+    function initApp() {
+      negotiatorAuthenticator.init().then(function (tokens) {
+        tokens.map(
+          ticketToken => {
+            const container = document.createElement("li");
```
colourful-land, line 36: Not needing `li` if you go `div` instead of `ul`

playground/index.html
```diff
@@ -0,0 +1,52 @@
...
+            const tokenId = document.createTextNode(ticketToken.id);
```
colourful-land, line 37: the member elements of a ticketToken are here (not `id`, `ticketId`)
https://github.com/AlphaWallet/blockchain-attestation/blob/master/data-model/SignedTicket.asn#L17-L19

playground/negotiator-authenticator.js
```diff
@@ -0,0 +1,172 @@
...
+    } catch (error) {
+        return [undefined, error];
+    }
+}
+
+// Mock of Negotiator
+const Negotiator = {
+    getTokenInstances: function async() {
+        return [
+            { id: 42, token: { ticketClass: "VIP" }, ownerAddress: null },
```
colourful-land, line 21: Oh, I see. You have the `id` outside of token object. Well actually it is inside because it needs to be signed (unlike `ownerAddress` which needn't to be signed. But it's up to you if you want to create an artificial `id` which is not the actual `ticetID`, just as long as the web developer know that they can't use this artificial `id` when communicating with the blockchain.

playground/negotiator-authenticator.js
```diff
@@ -0,0 +1,172 @@
...
+    // 2. Determine if the user has VIP tokens (for front end developers UI / UX)
+    let isVIP = false;
+    tokens.map(ticket => { if (ticket.token.ticketClass == "VIP") isVIP = true });
+    if (isVIP) document.getElementById("vip-only-section").style.display = "block";
+    // 3. Return the and tokens
+    return tokens;
+}
+
+// to handle the vip click event from the end user
+negotiatorAuthenticator.vipClickEvent = async () => {
```
colourful-land, line 55: My thought is Authenticator is the library module we provide for all tokens (e.g. Ether) hence doesn't have functions specific to a token (ticket)? So this part is the code written by a web developer, not a mock library module right?

playground/index.html
nicktaras: As per your initial design, I can update the container element to be an unordered list element. I think this makes sense from a semantics perspective also (where the tokens will be presented as a list).

<ul>
    <li>token</li>
</ul>

playground/index.html
nicktaras: I'll switch out id to ticketId.

playground/negotiator-authenticator.js
nicktaras: I'll bring all the attributes into the token as per example below (and remove the 'id' from the object):
{ token: { ticketId: 42, ticketClass: "VIP", conferenceId: 1 }, ownerAddress: null },

playground/negotiator-authenticator.js
nicktaras: I maybe confused here. 

"negotiatorAuthenticator.vipClickEvent" - do you mean this function should be moved to the client side html file for the web development team manage.

colourful-land: I mean:

- whatever library the TokenScript team provide to the web developers, it can't contain `vipClickEvent since it's too specific to the project the web developers are working on, therefore, negotiator-authenticator.js seem to be part of the web developers' project

- However, previously it returned mock data which should be returned by TokenScript's library (negotiator to be specific).

So the line between these 2 is not very clear to me: 𝑎) the code TokenScript libraries provides (or its mock), and 𝑏) the code web developers write

nicktaras: Thanks I see what you mean. 

I'll make some updates to split the logic to it's clear to see which parts of the code are for the web developer to implement and which are TokenScript.

Keeping the web developer code inside the index.html file for now, and the tokenScript code inside the negotiator-authenticator.js file.

nicktaras: @colourful-land, I've moved the logic to what I understand to be in the correct place. 
The developer code is in the index.html and the library code is inside the negotiator-authenticator.js file.

playground/index.html
```diff
@@ -0,0 +1,110 @@
...
+    function selectVipTicket(ticketId) {
+      event.stopPropagation();
+      // 1. Get most up to date Tokens (Confirm if this should be done each time)
+      negotiatorAuthenticator.init().then(function (tokens) {
+        // 2. Get the ticket using ID
+        const chosenTicket = tokens.filter(ticket => (ticket.token.ticketId == ticketId))[0];
+        // 3. Authenticator (non-Disney mode) - no owner - open modal to retrieve owner
+        if (chosenTicket.ownerAddress == null) {
+          // lead to email code modal process, created by Authenticator.
+          negotiatorAuthenticator.findOwner(chosenTicket);
```
nicktaras, line 87: 87: findOwner should be assertOwnerAddress() - is this correct? Thanks.

colourful-land: Yes that's right ('asset' is just from top of my head, anything stronger than 'find' will do)

.gitignore
```diff
@@ -1,4 +1,5 @@
 out
 .idea
 build
-.iml
\ No newline at end of file
+.iml
+.DS_Store
```
nicktaras, line 5: Added DS_Store to stop mac files reaching the repo.

playground/.gitignore
```diff
@@ -0,0 +1,2 @@
+# misc
+.DS_Store
```
nicktaras, line 2: Added DS_Store to stop mac files reaching the repo.

playground/index.html
```diff
@@ -0,0 +1,127 @@
...
+          negotiatorAuthenticator.assetOwnerAddress(chosenTicket);
+        }
+        // 4. If owner address is found, authenticateAddress
+        if (chosenTicket.ownerAddress == currentUser.ownerAddress) {
+          // this will lead to sign-message, even if user typed the code in email.
+
+          // TODO: Don’t encapsulate 'authenticateAddress' - use web3 challenge
+          let web3 = new Web3('HTTP://127.0.0.1:7545');
+          // Gnache / Mock Private Key only
+          const signature = web3.eth.accounts.sign('Some data', '35335e2eac26772c5b1aed0114f7f635b99eb17fe8ef37cd68edd7acf093fa9c');
```
nicktaras, line 101: Started to build logic to apply a web3 signature challenge before moving to next step. 

Questions:

1. How should the required control flow work at this point (e.g. should this hold a promise until resolved?)
2. What is the required user experience

playground/negotiator-authenticator.js
```diff
@@ -0,0 +1,154 @@
...
+        document.getElementById("modal").style.display = 'none';
+        alert('user can enter the VIP room');
+    }, function (error) {
+        // handle error
+    });
+}
+
+const negotiatorAuthenticator = {};
+
+negotiatorAuthenticator.init = async () => {
```
nicktaras, line 59: TODO - I will look at the naming used here and tidy it. Where if I am correct, we should only have Negotiator and Authenticator modules for the front end API.

colourful-land:
Closing this one so you can restart another against main (as we changed default branch)

PR #51, opened 2020-12-14
Added tests for magic link for ticket

jot2re:
Added tests to construct magic link for ticket


colourful-land (approved):
Thanks look nice to me

PR #52, opened 2020-12-16
Smaller attestation request

jot2re:
Changed Attestation request to only contain the user's public key instead of the full SPKI (since we always use ECDSA secp256k1).
This was suggested by Oleg to be consistent with what we store in e.g. Cheque and Ticket objects.


jot2re:
After discussion with Tomek we chose to not implement this due to complications for him. So I am deleting this request

PR #53, opened 2020-12-17
Updated to BN256

jot2re:
Updated the code to use BN256 group 1 instead of SECP256k1.
NOTE: testContract fails since the solidity is not updated. I think @JamesSmartCell has to do this.

Parameters for BN256 group 1 are as follows:
Field Size: 115792089237314936872688561244471742058375878355761205198700409522629664518163
Curve order: 115792089237314936872688561244471742058035595988840268584488757999429535617037
h (cofactor): 1
For equation y^2=x^3+ax+b:
a: 0
b: 3


colourful-land:
I'll leave this open until @JamesSmartCell updated the contract.

jot2re:
Still work in progress, since we need to change to Pedersen commitments

colourful-land:
Thanks a lot - closely monitoring progress on this as we originally set the delivery of issuer (ticket isssuer class ) by end of Dec.

jot2re:
I have updated this to use Pedersen commitments to prevent possible brute forcing of user ID when using pairing based elliptic curves for the math.

colourful-land (approved):
Builds fine…

PR #57, opened 2020-12-26
renaming fields to be more specific, and reorder them 

colourful-land:
I got someone confused and asked me why `conferenceId` when there is only one conference in Devcon (unlike some forum events which is made up of smaller conferences). The ordering and naming removes the confusion

### DIFF
--- a/data-module/SignedDevonTicket.asd
+++ b/data-module/SignedDevonTicket.asd
@@ -17,9 +17,9 @@
     <namedType name="Ticket">
         <type>
             <sequence>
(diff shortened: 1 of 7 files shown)


jot2re:
We can easily change this, but I must also adapt the code accordingly. Furthermore, we must also adapt the JS code as well and eventually the Solidity. So I will leave this PR open while we work on this.

jot2re:
I just updated the code to ensure that the encoded data fits the new format.

PR #58, opened 2020-12-26
fixed the test data directory

jot2re:
Closes #55

PR #59, opened 2020-12-26
made tests and demo deterministic

jot2re:
Made tests deterministic for simplicity.
Should be approved after #58 to ensure that everything gets merged into master


colourful-land (approved):
`AttestationCrypto` look like a utility class which shouldn't be passed around as parameter but we can do that refactoring later.

jot2re:
The issue is that not all methods are not static. Some require randomness. This need to be parsed to the class to allow for deterministic "randomness" for tests but proper pseudorandomness for execution. I think the alternative is to parse the randomness generator as an argument for the methods which require it, which might be better to do?

PR #60, opened 2020-12-27
Updated parameters

jot2re:
I updated the underlying field for the BN256 curve to match the one in [Ethereum](https://github.com/ethereum/go-ethereum/blob/master/crypto/bn256/cloudflare/constants.go) which turns out to be different from the ones recommended by [IETF](https://tools.ietf.org/id/draft-kasamatsu-bncurves-01.html#rfc.section.4.3.2) which I had originally used.


colourful-land (approved):
Thanks.

PR #61, opened 2020-12-27
Add integration tests (all test classes renamed!!!)

colourful-land:
so that default gradle build doesn't halt on `contractTest`

Now, gradle build works even of the smart contract doesn't. To trigger test with contract, do `gradle integrationTest`

there are only renaming stuff in the PR so you can safely approve it and merge if the two commands works correctly on your PC

`gradle build` should work
`gradle integrationTest` should fail with the testContract problem.
`gradle check` should fail too as it includes integrationTest.

See below.

````
$ gradle build

BUILD SUCCESSFUL in 1s
5 actionable tasks: 2 executed, 3 up-to-date
a@thinkpad:~/IdeaProjects/attestation$ gradle integrationTest

> Task :integrationTest

ProofOfKnolwedgeIntegTest > TestContract() FAILED
    org.opentest4j.AssertionFailedError at ProofOfKnolwedgeIntegTest.java:36

1 test completed, 1 failed

> Task :integrationTest FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':integrationTest'.
> There were failing tests. See the report at: file:///home/a/IdeaProjects/attestation/build/reports/tests/integrationTest/index.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 12s
3 actionable tasks: 2 executed, 1 up-to-date
````

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,11 @@ repositories {
     mavenCentral()
 }
 
+configurations {
+    intTestImplementation.extendsFrom implementation
+    intTestRuntimeOnly.extendsFrom runtimeOnly
+}
+
 dependencies {
     implementation 'org.apache.logging.log4j:log4j-api:2.13.0'
     implementation 'org.apache.logging.log4j:log4j-core:2.13.0'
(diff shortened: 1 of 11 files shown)


JamesSmartCell (approved):

jot2re:
`gradle integrationTest` and `gradle check` fails as expected, but `gradle build` also fails as well for me

jot2re (changes requested):
`gradle integrationTest` and `gradle check` fails as expected, but `gradle build` also fails as well for me

colourful-land:
> `gradle integrationTest` and `gradle check` fails as expected, but `gradle build` also fails as well for me

I mistook `check` as a higher level task, but it is a lower level task than build hence build covers it.  Now the expected behaviour is:

`gradle build` should work
`gradle integrationTest` should fail with the testContract problem.

Make sure to update the code - it is rebased - then test

jot2re (approved):
This seems to fix the issue

PR #62, opened 2020-12-27
Nodejs in gradle configured to use Bhara's code

colourful-land:
Solves issue #48 

Please verify it is working on your computer. On my laptop which already have nodejs v15 installed, it works like this:

````
$ gradle testJavaScript
Configuration on demand is an incubating feature.

> Task :compileJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :compileTestJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :testJavaScript
devconId = 6, ticketId = 546048445646851568430134455064804806, ticketClass = 0, riddle = 55588602291676239330812044804329807443759080566389442831624630380667862134310732664343459533540896792011893045114015709379942436306788761339283410175493380

BUILD SUCCESSFUL in 14s
4 actionable tasks: 4 executed
````

If it doesn't work for you on Windows, it's maybe because nodejs isn't made available to gradle. You can make sure node is in the %PATH% environment variable or work on editing the gradle configuration file, instruction here https://github.com/colourful-land/gradle-node-plugin/blob/master/docs/usage.md (watch for the `download = false` part)

PR #66, opened 2020-12-30
Refactor dataobject class fixes #64

### LINKED ISSUE #64: adjust the SignedDevconTicket paser following 3-layer concept
https://github.com/TokenScript/attestation/tree/3-layer-js-concept/3-layers leads to the 3 changes:

1. dataobject classes (which is the parser/serialiser) should be honest to the data structure and not be token objects (which is one layer above). That is, `SignedDevconTicket.ticket.ticketId`, not `SignedDevconTicket.ticketId`. It's Olegh's job to wrap that data and give the web developer Webster the tokenObject which has `tokenObject.ticketId`
2. dataObject classes can instantiated by a DER ArrayBuffer
3. (moved to another issue #65 ): test-cases should be expanded to cover both case: that the file created by Java test cases should be parsed correctly to dataObject and if you feeda dataObject to instantiate another one the result should serialise to the identical file.

colourful-land:
This is my test output so the structure is like specified in the 3-layer concept hence fixes #64

````
$ gradle testJavaScript
Starting a Gradle Daemon, 1 stopped Daemon could not be reused, use --status for details

> Task :testJavaScript
SignedDevconTicket {
  ticket: DevconTicket {
    devconId: 6,
    ticketId: 546048445646851568430134455064804806n,
    ticketClass: 0,
    riddle: 55588602291676239330812044804329807443759080566389442831624630380667862134310732664343459533540896792011893045114015709379942436306788761339283410175493380n
  },
  signatureAlgorithm: AlgorithmIdentifier { algorithmId: '' },
  signatureValue: ArrayBuffer {
    [Uint8Contents]: <30 44 02 20 47 8d f4 b1 35 4c 4f 10 d5 24 d9 82 f6 e9 c9 ff fe af 98 b4 ed 61 04 b4 16 ac 8e d9 b4 90 1d 88 02 20 27 3e 7a ff 5a b6 af 99 2d 4d ac 08 2b a1 c9 eb 61 e9 9f 4c 29 47 11 54 f4 72 3c f9 64 fc c5 f5>,
    byteLength: 70
  }
}

BUILD SUCCESSFUL in 5s
6 actionable tasks: 1 executed, 5 up-to-date
````

### DIFF
--- a/src/main/javascript/AlgorithmIdentifier.js
+++ b/src/main/javascript/AlgorithmIdentifier.js
@@ -8,37 +8,39 @@ export default class AlgorithmIdentifier {
   //**********************************************************************************
   /**
    * Constructor for AlgorithmIdentifier class
-   * @param {Object} [parameters={}]
-   * @param {Object} [parameters.schema] asn1js parsed value to initialize the class from
+   * @param {Object} [source={}]
+   * @param {Object} [source.schema] asn1js parsed value to initialize the class from
    * @property {string} [algorithmId] ObjectIdentifier for algorithm (string representation)
    */
-  constructor(parameters = {}) {
-    //region Internal properties of the object
-    /**
-     * @type {string}
-     * @desc ObjectIdentifier for algorithm (string representation)
-     */
-    this.algorithmId = getParametersValue(
-      parameters,
-      "algorithmId",
-      AlgorithmIdentifier.defaultValues("algorithmId")
-    );
-
-    if ("algorithmParams" in parameters)
+  constructor(source = {}) {
+    if (typeof(source) == "string") {
+      throw new TypeError("Not accepting string. For base64, convert to ArrayBuffer.")
+    }
+    if (source instanceof ArrayBuffer) {
+      const asn1 = fromBER(source)
+      this.fromSchema(asn1.result);
+    } else {
       /**
-       * @type {Object}
-       * @desc Any algorithm parameters
+       * @type {string}
+       * @desc ObjectIdentifier for algorithm (string representation)
        */
-      this.algorithmParams = getParametersValue(
-        parameters,
-        "algorithmParams",
-        AlgorithmIdentifier.defaultValues("algorithmParams")
+      this.algorithmId = getParametersValue(
+          source,
+          "algorithmId",
+          AlgorithmIdentifier.defaultValues("algorithmId")
       );
-    //endregion
 
-    //region If input argument array contains "schema" for this object
-    if ("schema" in parameters) this.fromSchema(parameters.schema);
-    //endregion
+      if ("algorithmParams" in source)
+        /**
+         * @type {Object}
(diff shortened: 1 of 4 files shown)


darakhbharat (commented):
Hi Weiwu,

I have tested the new code one difference I found is now character 'n' is getting appended for ticketId and riddle which was not the case before.

One more obervation is When I uploaded/tested - build/test-results/signed-devcon-ticket.der on https://lapo.it/asn1js/ the result is not matching 100%, see the riddle value is different and also there are more extracted values. Might not be an issue just added my obervation.

Where is the input data of that der file? Intially I was refering TestUseTicket.java but there are changes in ticket object now like conferenceId is replaced with devconId.

due to above obervations wanted to test it for multiple inputs to confirm hence not approved the PR. But if you think its not an issue then we can go ahead.

colourful-land:
> Hi Weiwu,
> 
> I have tested the new code one difference I found is now character 'n' is getting appended for ticketId and riddle which was not the case before.

`n` just mean bigInt. `riddle` was already bigInt in your code:
````
    if ("riddle" in asn1.result){
      const hex = bufferToHexCodes(
          asn1.result["riddle"].valueBlock.valueHex
      );

      this.riddle = BigInt(`0x${hex}`);
    }
````
except you don't see them as you converted them to String before printing. But I do notice the inconsistency (why ticket is not BigInt? and will need to file an issue outside of this PR to discussion how to deal with it.

> One more obervation is When I uploaded/tested - build/test-results/signed-devcon-ticket.der on https://lapo.it/asn1js/ the result is not matching 100%, see the riddle value is different and also there are more extracted values. Might not be an issue just added my observation.

Interesting observation! I thought the message was deterministic and that's what I told ligi (devcon lead) - now I have to update him. I think the Java test code creates a riddle differently each time. Will catch up with ligi on that (devcon).

> Where is the input data of that der file? Intially I was refering TestUseTicket.java but there are changes in ticket object now like conferenceId is replaced with devconId.

It was created by a test case `src/test/java/com/alphawallet/attestation/ticket/TestTicket.java` so all inputs are in that file (or generated from there). There were 2 clues: first you can search for the keyword `signed-devcon-ticket.der` to find which file has it; or it is listed in the files changed in the last PR https://github.com/TokenScript/attestation/pull/62 - the change of data module (conferenceId -> devconID) happened a while ago.

> due to above obervations wanted to test it for multiple inputs to confirm hence not approved the PR. But if you think its not an issue then we can go ahead.

Thanks for the careful review! We can go ahead since these observations didn't lead to bugs.

colourful-land:
Okay, the riddle mismatch problem is solved with a new commit. Other observations you made are issues outside of this PR and will be filed separately

darakhbharat (approved):
As discussed for now we are good with this PR and can reopen the issue in future if required.

PR #67, opened 2020-12-30
Added attestation request asn module, removed dead modules and fixed …

jot2re:
…some consistency issues in the rest


colourful-land:
You may as well replace all 'riddle' with a better name such as Pedersen Commit (and update the Java source code too - I will update the JavaScript source code accordingly.

colourful-land (approved):

PR #68, opened 2020-12-30
Deleted the obsolete version of the attesation protocol which require…

jot2re:
…d the commercial framework by objsys.
If we want to be "pure" we should also delete the obsolete and relatively irrelevant md files data.md, encoding.md, implementation.md from data-modules


colourful-land (approved):
You are right and I moved Implementation.md data.md encoding.md
into a branch in the document repo to update later according to the code:
https://github.com/TokenScript/documents/tree/attestation-code-doc-sync

PR #73, opened 2020-12-31
Update contract for bn256

JamesSmartCell:
Update contract for BN256.

Note: I couldn't work out the parameters for the Barrett reduction algorithm, there are modulus factors which aren't clear what values to pick. Namely, the value for the Barret_mod, which each description gives a seemingly different way to calculate it. There's no solid, stand-alone implementation I could find - all of them I saw had dependencies on other math libraries, which isn't a good sign for working on Ethereum.

The current implementation is a common blockchain hack, where you hash the value twice. I notice the value can be negative due to Two's complement interpretation but that's easy to fix.

However, if you could advise the correct values for the constants in the reduction we can switch back to that method.

### DIFF
--- a/src/test/java/com/alphawallet/attestation/ProofOfKnowledgeTest.java
+++ b/src/test/java/com/alphawallet/attestation/ProofOfKnowledgeTest.java
@@ -103,4 +103,22 @@ public class ProofOfKnowledgeTest {
     assertFalse(crypto.verifyEqualityProof(com1, com2, newPok));
   }
 
+  @Test
+  public void TestContract() {
+    SmartContract sc = new SmartContract();
+
+    for (int i = 0; i < 30; i++) {
+      ProofOfExponent pok = crypto.computeAttestationProof(BigInteger.TEN);
+      assertTrue(crypto.verifyAttestationRequestProof(pok));
+      assertTrue(sc.testEncoding(pok));
+    }
+
+    //now check fail
+    for (int i = 0; i < 5; i++) {
+      ProofOfExponent pok = crypto.computeAttestationProof(BigInteger.TEN);
+      assertTrue(crypto.verifyAttestationRequestProof(pok));
+      ProofOfExponent newPok = new ProofOfExponent(pok.getBase(), pok.getRiddle(), pok.getPoint(), pok.getChallenge().add(BigInteger.ONE));
+      assertFalse(sc.testEncoding(newPok));
+    }
+  }
 }
--- a/src/test/java/com/alphawallet/attestation/demo/AttestationValidityCheck.sol
+++ b/src/test/java/com/alphawallet/attestation/demo/AttestationValidityCheck.sol
@@ -1,0 +1,497 @@
+/* Attestation decode and validation */
+/* AlphaWallet 2020 */
+
+pragma solidity ^0.6.0;
+
+contract DerDecode {
+    address payable owner;
+
+    bytes1 constant BOOLEAN_TAG         = bytes1(0x01);
+    bytes1 constant INTEGER_TAG         = bytes1(0x02);
+    bytes1 constant BIT_STRING_TAG      = bytes1(0x03);
+    bytes1 constant OCTET_STRING_TAG    = bytes1(0x04);
+    bytes1 constant NULL_TAG            = bytes1(0x05);
+    bytes1 constant OBJECT_IDENTIFIER_TAG = bytes1(0x06);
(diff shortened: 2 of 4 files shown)


src/test/java/com/alphawallet/attestation/ProofOfKnowledgeTest.java
```diff
@@ -103,4 +103,22 @@ public void TestNegativeEqualityProof() {
     assertFalse(crypto.verifyEqualityProof(com1, com2, newPok));
   }
 
+  @Test
+  public void TestContract() {
+    SmartContract sc = new SmartContract();
+
+    for (int i = 0; i < 30; i++) {
+      ProofOfExponent pok = crypto.computeAttestationProof(BigInteger.TEN);
+      assertTrue(crypto.verifyAttestationRequestProof(pok));
```
jot2re, line 112: Furthermore, for the smart contract we should be using the proof of equality, i.e. crypto.computeEqualityProof

src/test/java/com/alphawallet/attestation/demo/AttestationValidityCheck.sol
```diff
@@ -0,0 +1,497 @@
...
+        pok.base = octet.byteCode;
+        octet = decodeOctetString(attestation, octet.decodeIndex);
+        pok.riddle = octet.byteCode;
+        octet = decodeOctetString(attestation, octet.decodeIndex);
+        pok.challenge = bytesToUint(octet.byteCode);
+        octet = decodeOctetString(attestation, octet.decodeIndex);
+        pok.tPoint = octet.byteCode;
+
+        //The hand optimised concat4 is more optimal than using abi.encodePacked (checked the gas costs of both)
+        bytes memory cArray = concat4(GPoint, pok.base, pok.riddle, pok.tPoint);
```
jot2re, line 100: The values of the two commitments used in the equality check should also be included in cArray

src/test/java/com/alphawallet/attestation/demo/AttestationValidityCheck.sol
```diff
@@ -0,0 +1,497 @@
...
+        OctetString memory octet;
+
+        require(attestation[0] == (CONSTRUCTED_TAG | SEQUENCE_TAG));
+
+        len = decodeLength(attestation, 1);
+        octet.decodeIndex = len.decodeIndex;
+
+        //decode parts
+        octet = decodeOctetString(attestation, octet.decodeIndex);
+        pok.base = octet.byteCode;
```
jot2re, line 91: It should be ensured that pok.base is the H point

src/test/java/com/alphawallet/attestation/demo/AttestationValidityCheck.sol
```diff
@@ -0,0 +1,497 @@
...
+        octet = decodeOctetString(attestation, octet.decodeIndex);
+        pok.riddle = octet.byteCode;
```
jot2re, line 93: The riddle should actually be computed from the commitment in the Ticket and in the Attestation, see AttestationCrypto.verifyEqualityProof

src/test/java/com/alphawallet/attestation/demo/AttestationValidityCheck.sol
```diff
@@ -0,0 +1,497 @@
...
+        }
+    }
+
+    //TODO: Implement Barrett reduction for mod512 operation
+    function mapToInteger(bytes memory input) private pure returns(uint256)
+    {
+        bytes32 idHash = keccak256(abi.encodePacked(uint8(0x00), input));
+        bytes32 idHash2 = keccak256(abi.encodePacked(uint8(0x01), idHash));
+        int256 idNum = int256(bytes32ToUint(idHash2));
+        idNum = idNum % int256(curveOrder);
```
jot2re, line 215: Like for the java code, this is unfortunately not enough

jot2re (changes requested):
NOTE: I have not checked the ASN decoding aspects of the code. Let me know if you also want me to check this.

In regards to the crypto part I have some minor comments which I have described in-line in the code. The biggest issue remaining is the price of doing modulo on an integer larger than 256 bits. As we talked about on Discord even using Barrett's reduction we still need to multiply two 256 bit numbers together, which will already be expensive. However, maybe we could do this using the base-step of Karatsuba to only use 4 regular 256 bit multiplications?
In regards to the smart contract itself it looks like none of the non-cryptographic checks on the Attestation and Ticket is happening, are these checks supposed to be part of this PR or can we wait with those until a later PR?

PR #74, opened 2020-12-31
Handled issue 70

### LINKED ISSUE #70: Reduce transaction size: by not inclduing the EC parameter or include in a condensed way
Many bytes are wasted to describe 0 and 7 (byte 150 and byte 184)

````
$ openssl asn1parse -inform DER < build/test-results/signed-devcon-ticket.der
    0:d=0  hl=4 l= 392 cons: SEQUENCE          
    4:d=1  hl=2 l=  78 cons: SEQUENCE          
    6:d=2  hl=2 l=   1 prim: INTEGER           :06
    9:d=2  hl=2 l=   3 prim: INTEGER           :BE06
   14:d=2  hl=2 l=   1 prim: INTEGER           :00
   17:d=2  hl=2 l=  65 prim: OCTET STRING      [HEX DUMP]:040AFDF0E2474CAE9B6616F04BACDD9F76AB5882DBB8399D3F60A15361DAD7030F27BE3F583FE45DE9495F84F48237EC2B7C710EB3B5D2E7A5652D8D56C718256F
   84:d=1  hl=3 l= 236 cons: SEQUENCE          
   87:d=2  hl=2 l=   7 prim: OBJECT            :id-ecPublicKey
   96:d=2  hl=3 l= 224 cons: SEQUENCE          
   99:d=3  hl=2 l=   1 prim: INTEGER           :01
  102:d=3  hl=2 l=  44 cons: SEQUENCE          
  104:d=4  hl=2 l=   7 prim: OBJECT            :prime-field
  113:d=4  hl=2 l=  33 prim: INTEGER           :FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F
  148:d=3  hl=2 l=  68 cons: SEQUENCE          
  150:d=4  hl=2 l=  32 prim: OCTET STRING      [HEX DUMP]:0000000000000000000000000000000000000000000000000000000000000000
  184:d=4  hl=2 l=  32 prim: OCTET STRING      [HEX DUMP]:0000000000000000000000000000000000000000000000000000000000000007
  218:d=3  hl=2 l=  65 prim: OCTET STRING      [HEX DUMP]:0479BE667EF9DCBBAC55A06295CE870B07029BFCDB2DCE28D959F2815B16F81798483ADA7726A3C4655DA4FBFC0E1108A8FD17B448A68554199C47D08FFB10D4B8
  285:d=3  hl=2 l=  33 prim: INTEGER           :FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141
  320:d=3  hl=2 l=   1 prim: INTEGER           :01
  323:d=1  hl=2 l=  71 prim: BIT STRING        
````

jot2re:
I refactored the signed ticket object.
This closes #70 at least in the java code.
@colourful-land are you handling the JS code or should I notify Oleg and Bharat?

### DIFF
--- a/data-modules/output/asn/SignedCheque.asn
+++ b/data-modules/output/asn/SignedCheque.asn
@@ -16,7 +16,7 @@ SignedCheque ::= SEQUENCE {
 Cheque ::= SEQUENCE {
     amount         INTEGER,
     validity       ValidityValue,
-    riddle         OCTET STRING 
+    commitment     OCTET STRING
 }
 		
 
--- a/data-modules/output/asn/SignedDevconTicket.asn
+++ b/data-modules/output/asn/SignedDevconTicket.asn
@@ -7,19 +7,25 @@ IMPORTS
 AlgorithmIdentifier FROM AuthenticationFramework;
 
 SignedDevconTicket ::= SEQUENCE {
-    ticket               DevconTicket,
-    signatureAlgorithm   AlgorithmIdentifier,
-    signatureValue       BIT STRING
+    ticket           DevconTicket,
+    commitment       OCTET STRING,
+--  The algorithm and public key are optional since they will normally be internally defined from devconId --
+    publicKeyInfo    PublicKeyInfo OPTIONAL,
+    signatureValue   BIT STRING
 }
 
 
 DevconTicket ::= SEQUENCE {
-    devconId     INTEGER,
+    devconId         INTEGER,
     ticketId         INTEGER,
     ticketClass      INTEGER,
 --  (currently not specified)      --
 --  co2_token         OCTET STRING --
-    riddle         OCTET STRING
+}
+
+PublicKeyInfo ::= SEQUENCE {
+    algorithm       AlgorithmIdentifier,
+    publicKey       BIT STRING
 }
 
 END
--- a/data-modules/src/SignedCheque.asd
+++ b/data-modules/src/SignedCheque.asd
@@ -20,8 +20,8 @@
             <sequence>
                 <element name="amount" type="asnx:INTEGER"/>
                 <element name="validity" type="ValidityValue"/>
-                <element name="riddle" type="asnx:OCTET-STRING">
-                    <annotation>The elliptic curve point that is the riddle</annotation>
+                <element name="commitment" type="asnx:OCTET-STRING">
+                    <annotation>The elliptic curve point that is the commitment to the user's identifier</annotation>
                 </element>
             </sequence>
         </type>
--- a/data-modules/src/SignedDevconTicket.asd
+++ b/data-modules/src/SignedDevconTicket.asd
@@ -8,9 +8,14 @@
                 <element name="ticket" type="DevconTicket">
                     <annotation>The actual, unsigned, ticket object</annotation>
                 </element>
-                <element name="signatureAlgorithm" type="AlgorithmIdentifier"/>
-                <element name="signatureValue" type="asnx:BIT-STRING">
+                <element name="commitment" type="asnx:OCTET-STRING">
+                  <annotation>The elliptic curve point that is a commitment to the ticket holder's identity</annotation>
                 </element>
+                <!-- The algorithm and public key are optional since they will normally be internally defined from devconId -->
+                <optional>
+                  <element name="publicKeyInfo" type="PublicKeyInfo"/>
+                </optional>
+                <element name="signatureValue" type="asnx:BIT-STRING"/>
             </sequence>
         </type>
     </namedType>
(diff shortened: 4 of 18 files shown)


colourful-land:
@darakhbharat are you available to update JS following the data module update?

colourful-land:
> @darakhbharat are you available to update JS following the data module update?

I patched up the JavaScript side. Now the problem is with the Solidity test code.

colourful-land (commented):
Having fixed the JavaScript part I realised this issue is more complicated then I thought.

First, there is no `ec_verify` in Solidity - Vitalik has been creative and thought you only needed ec_recover hence saving all the bytes used to store the public keys. Someone did [an `ec_verify`](https://gist.github.com/axic/5b33912c6f61ae6fd96d6c4a47afde6d) but since it uses ec_recover under the hood, it uses more resources than ec_recover. 

That means any data prepared for the consumption of Ethereum has no benefit - only drawback - to include PublicKey. This leads to 2 actions.

1. Find out if we can encode the V bit (Vitalik bit, needed for `ec_recover`) somewhere.
2. Drop the public key completely from the data structure.

## Regarding 1: Find out if we can encode the V bit

I verified that currently signature in the BITSTRING is stored as a sequence

ECDSA-Sig-Value ::= SEQUENCE { r INTEGER, s INTEGER }

````
$ echo 30 44 02 20 38 db 21 b6 b5 b7 c6 92 da ad a2 b6 2e bb 89 e5 a3 6e 3a 3c ce 66 1e 38 53 2b c9 ac c8 5c 34 1b 02 20 70 46 73 21 8f 77 b2 47 b5 51 ab 3c 3d 74 e1 ef 8f 4f 7e 3a e0 40 1d 53 54 26 65 3a aa 5e c2 a2 | xxd -r -p |  openssl asn1parse -inform DER 
    0:d=0  hl=2 l=  68 cons: SEQUENCE          
    2:d=1  hl=2 l=  32 prim: INTEGER           :38DB21B6B5B7C692DAADA2B62EBB89E5A36E3A3CCE661E38532BC9ACC85C341B
   36:d=1  hl=2 l=  32 prim: INTEGER           :704673218F77B247B551AB3C3D74E1EF8F4F7E3AE0401D535426653AAA5EC2A2
````

If we add a V-value as the 3rd member of the sequence, will this be considred blasphemy or create more problems down the way?

> If you are interested, the Etheruem way is to pass 𝑟, 𝑠, 𝑣 as separate parameters to a smart contract, but that wouldn't work for us when we have a data structure that has a lot of stuff in it: more than 1 signatures, some zk proof etc, easily bloat the contract parameters into something like this:

>      contract.vote(candidate, message, commitment, proof, sig1_r, sig1_s, sig1_v, sig2_r sig2_s, sig2_v)

## Regarding 2: Drop the public key completely from the data structure.

One might think the public key is useful when the client wishes to verify a ticket, but 1) the standard client web3.js also does not support `ec_verify`, again courtesy of Vitalik. introducing other client libraries increases complexity; 2) the only one benefit from that is perhaps openssl based systems such as PHP, but, thanks to the popularity of Etheruem, even PHP started to have an ec_recovery library, as Oleg pointed out.

Furthermore, the current public key is either hard-coded or managed by a smart contract (in reality smart contract today only returns the first 20 bytes of the hash of the public key since you can recover a public key yourself, see point 1), so there is no point for a client to get it from the `SignedDevconTicket` object.

colourful-land (changes requested):
I proposed two actions to be taken, first about Vitalik value, second about the omission of PublicKeyInfo. To make it quick, let's approve this once we have a solution for V value; as of publicKeyInfo you can just omit it in the unit test output and the refactoring that touches ASN.1 module can be done later in a separate PR.

jot2re:
@colourful-land 
Regarding suggestion 1; I think changing the format of a standard object like ECDSA-Sig-Value, by adding an extra field is not a good idea. However, as far as I understand the purpose of the v value is to contain a single bit describing the "sign" of the y coordinate of the signature. I am not sure exactly which value of v is used to represent what, but I assume the value 27 is used to represent the smallest valid value of y and 28 the largest valid value. Maybe @JamesSmartCell can confirm this? 
In any case we can simply just ensure that the smallest valid y valid is always used. This means that the v value is basically hard-coded and does not need to be included.

Regarding suggestion 2; I think you might be right that it *can* make sense to leave out the public key info and assume the websites have that info hardcoded for the specific key used for a specific DevCon. However, in general it is more future-proof and robust to add the public key info to an object, because in systems like x509 if a key gets compromised it will be added to a revocation list and thus never be accepted by a website. Thus preventing malicious use of a compromised/obsolete key for a future DevCon. Still since we are in the situation where the websites must support this very specific protocol, expecting signatures from a single, or very limited amount of keys, it can make sense to not include the public key info as I assume any potential revocation will have to be hard-coded. Still, I do believe keeping the public key info optional is more future proof.

colourful-land (changes requested):
Okay agreed. Normally I should wait for @JamesSmartCell to update his contract so that DER decoder doesn't throw up at the changed SignedDevconTicket data object, but knowing his working style (of not having too many small PRs), I think he will want to do this in the pending PR Update contract for bn256 https://github.com/TokenScript/attestation/pull/73 so I take the liberty of approving this and letting this one in.

colourful-land (approved):
This fixes #70 as well as fixes #76

PR #79, opened 2021-01-02
Refactored the magic link tests

### LINKED ISSUE #78: add MagicLink in the test case
Specifically, it's needed in TestTicket so the devcon folks can see how they can generate one of those.

jot2re:
This closes #78


colourful-land (approved):
Thanks.

Issue #77, opened 2021-01-02
need a solution for 𝑣 value in smart contract


colourful-land:
When a website wishes to authenticate a decon ticket, the wallet / javascript may submit a data object `SigendDevconTicket` with `PublicKeyInfo`.

But `PublicKeyInfo` is expected to be left out in transaction since Solidity doesn't support ec_verify anyway (@JamesSmartCell or @James-Sangalli know whether the assumption is true?), there is no gas saving by attaching the public key. Therefore, the solidity counteract need to figure out the 𝑣 value to use when calling ec_recover.

Quote from @jot2re :

> Regarding suggestion 1; I think changing the format of a standard object like ECDSA-Sig-Value, by adding an extra field is not a good idea. However, as far as I understand the purpose of the v value is to contain a single bit describing the "sign" of the y coordinate of the signature. I am not sure exactly which value of v is used to represent what, but I assume the value 27 is used to represent the smallest valid value of y and 28 the largest valid value. Maybe @JamesSmartCell can confirm this?
> In any case we can simply just ensure that the smallest valid y valid is always used. This means that the v value is basically hard-coded and does not need to be included.

colourful-land:
What is actionable for @jot2re's suggestion to be implemented? Should we modify the Devcon isssuer class? in that case this issue should be assigned to @jot2re

jot2re:
Actually the code already ensures that the smallest value of y is used. So I think this is actually just an issue of whether we can use this to hardcode the v value in Solidity. I think @JamesSmartCell is the one who would know this

colourful-land:
> So I think this is actually just an issue of whether we try both possible values of v or if we can put constraints on the signing key being generated.

Okay, I'm in preference of trying both value v in the smart contract. The smart contract should have an idea what public key to expect (hard-coded or, even better, returned from a smart contract, but perhaps in the next a few months though).

JamesSmartCell:
The important thing with 𝑣 is only if it's odd or even. It's a convenience for calculating the Public Key from the signature. If even, then we take the point above the x-axis, if odd then the one below. It's recommended to only use v=27 and v=28 although I've seen dapps using 0 and 1. If we control all the flow paths then we know it's either 27 or 28, if not, then we need to check odd or even.

jot2re:
I realised that we actually don't even need to generate the public key to have a specific sign. We can instead simply store the *v* value with the public key address we need to verify against. In that way we only need to do one ecRecover.

[...]

PR #81, opened 2021-01-03
Ensure that generated ec keys always uses the smallest possible y value

jot2re:
Addresses issue #77 
I changed code to always generate ecdsa keys with the lowest of the two possible y coordinates for the public key. This is tested in CryptoTest.testECKeyWithLowY


colourful-land (approved):
I took the liberty of refactoring the change into a class so the original class is untouched.

PR #83, opened 2021-01-04
 the class needed as a commandline, for devcon

colourful-land:
hi @jot2re I need a bit of help here.

First, read the document of our deliverable which I posted to the general channel. you only need to check the first page where the JavaClass is described.

Then check this half done code - the thing I'm stuck is how to create a Ticket object from a privatekey instead of `ECPrivateKeyParameters`:

https://github.com/TokenScript/attestation/blob/cd719062169c9c55c1e4670cf4944fb0e8f1f521/src/main/java/org/devcon/ticket/Issuer.java#L44

To test the code in run time you need to 

````
$ gradle shadowJar

BUILD SUCCESSFUL in 3s
2 actionable tasks: 2 executed
$ java -cp ./build/libs/attestation-all.jar org.devcon.ticket.Issuer /tmp/file.pem 2 3 4 5
````

Where the needed `/tmp/file.pem` can be generated from the instruction in the deliverable document.

### DIFF
--- a/src/main/java/com/alphawallet/attestation/core/DERUtility.java
+++ b/src/main/java/com/alphawallet/attestation/core/DERUtility.java
@@ -1,23 +1,24 @@
 package com.alphawallet.attestation.core;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.math.BigInteger;
-import java.util.Arrays;
-import java.util.List;
-import org.bouncycastle.asn1.ASN1EncodableVector;
-import org.bouncycastle.asn1.ASN1InputStream;
-import org.bouncycastle.asn1.ASN1OctetString;
-import org.bouncycastle.asn1.ASN1Sequence;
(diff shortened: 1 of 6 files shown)


jot2re:
Is it strictly needed to read it as a JDK PrivateKey first? If not, I already wrote a helper method for loading keys from disc in connection with the Demo class. Basically the method DERUtility.restoreBase64Keys() reads a PEM file with a private EC key and returns a AsymmetricCipherKeyPair that is needed in Ticket

colourful-land:
> Is it strictly needed to read it as a JDK PrivateKey first?

No. I am confused of which one is the RFC private key which one is the JDK and it seems I used the wrong one.

>  If not, I already wrote a helper method for loading keys from disc in connection with the Demo class. Basically the method DERUtility.restoreBase64Keys() reads a PEM file with a private EC key and returns a AsymmetricCipherKeyPair that is needed in Ticket.
> I have committed a change that uses this method.

Great!

Issue #72, opened 2020-12-31
Reduce transaction size: not including Pedersen commit in the ProofOfExponent


colourful-land:
Is it possible?

jot2re:
This is closed by PR #89

PR #87, opened 2021-01-05
Refactor contract for optimum gas usage and add verifyEqualityProof

JamesSmartCell:
- add verifyEqualityProof to contract
- add corresponding positive and negative soak tests


src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,793 @@
...
+        pok.base = octet.byteCode;
+        octet = decodeOctetString(attestation, octet.decodeIndex);
+        pok.riddle = octet.byteCode;
+        octet = decodeOctetString(attestation, octet.decodeIndex);
+        pok.challenge = bytesToUint(octet.byteCode);
+        octet = decodeOctetString(attestation, octet.decodeIndex);
+        pok.tPoint = octet.byteCode;
+
+        //Calculate LHS ECPoint
+        (uint256 x, uint256 y) = extractXYFromPoint(pok.base);
```
jot2re, line 225: It should also be checked that the base is H.

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,793 @@
...
+        
+        x = uint256(xBytes);
+        y = uint256(yBytes);
+    }
+
+    function mapToCurveMultiplier(bytes memory input) public pure returns(uint256 res)
+    {
+        bytes memory nextInput = input;
+        do
+        {
```
jot2re, line 333: I will make the change in the rejection-sampling branch to ensure that `uint256(idHash) >> curveOrderBitShift` is less than curve order. Thus the loop will not be needed, but only a check that `uint256(idHash) >> curveOrderBitShift` is in fact less than curve order. If not the client has been malicious and the request should fail.

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,793 @@
...
+        if(msg.sender == owner)
+        {
+            selfdestruct(owner);
+        }
+        else revert();
+    }
+    
+    //Most optimal 4 byte array variable length concatenate. If sizes of the inputs are static we can optimise this even further
+    function concat4(
+        bytes memory _bytes1,
```
jot2re, line 531: These are in fact static since they are all elliptic curve points. So maybe we can optimise a bit more here?

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,793 @@
...
+            mstore(0x40, and(
+            add(add(end, iszero(end)), 31),
+            not(31) // Round down to the nearest 32 bytes.
+            ))
+        }
+
+        return tempBytes;
+    }
+    
+    function concat5(
```
jot2re, line 656: Like for concat4, these are all also of constant size

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,793 @@
...
+        
+        //Ensure base is correct
+        if (!ecEquals(base, H))
+        {
+            revert();
+        }
+        
+        //Note: if needed we can hand optimise this, but the gain is small
+        //bytes memory cArray = abi.encodePacked(GPoint, pok.base, comPoint1, comPoint2, pok.tPoint); //
+        bytes memory cArray = concat5(GPoint, pok.base, com1, com2, pok.tPoint);
```
jot2re, line 170: We strictly don't need the G point. It can be removed and will be as part of issue #72.
It is also sufficient to only hash the X coordinate and a bit describing the sign of the y coordinate. Although this would require the computation of the sign which requires a comparison to curveOrder. So I am not sure if that is better?

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,793 @@
...
+        (uint256 x, uint256 y) = extractXYFromPoint(pok.base);
+        uint256[2] memory lhs = ecMul(pok.challenge, x, y);
+        
+        if (lhs[0] == 0 && lhs[1] == 0) { return false; } //early revert to avoid spending more gas
+
+        //Calculate RHS ECPoint
+        (x, y) = extractXYFromPoint(pok.riddle);
+        
+        //The hand optimised concat4 is more optimal than using abi.encodePacked (checked the gas costs of both)
+        bytes memory cArray = concat4(GPoint, pok.base, pok.riddle, pok.tPoint);
```
jot2re, line 234: We strictly don't need the G point. It can be removed and will be as part of issue #72

jot2re (changes requested):
It looks good. I added some minor comments and ideas for optimisations. 
One major thing (in relation to deployment) is that the validity of the ticket and its signature and the signature of the attestation is currently not verified. Can you please confirm that this should not be included in this PR?

jot2re:
BTW please cross-reference with the updates I made to PR #88

PR #88, opened 2021-01-05
Rejection sampling

jot2re:
I implemented rejection sampling to ensure that when a user computes a ZK proof, he/she has already done rejection sampling when picking randomness to ensure that the contract can simply use KECCAK256 to restore the challenge value used in the proof.
NOTE: That as a side effect the generator points have also been updated (they don't need to, but I much prefer to have a test in the code base that computes these using our production code to prove to anyone that they have not been constructed maliciously).
However, because of this we also need Oleg to make an update to the JS.


oleggrib:
I updated my JS code according to the JAVA changes. commit pushed.

colourful-land:
> I updated my JS code according to the JAVA changes. commit pushed.

Very impressive! Thanks.

PR #91, opened 2021-01-06
Compressing transaction contract

JamesSmartCell:
Update the contract to reflect latest cryptography.

Also I removed the smart contract test in the Demo class test. Currently it won't work but let's work out what is meant to happen here and add it.


src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,633 @@
...
+    event RtnS(string val);
+
+    uint256[2] private G = [ 15729599519504045482191519010597390184315499143087863467258091083496429125073,
+    1368880882406055711853124887741765079727455879193744504977106900552137574951 ];
+
+    uint256[2] private H = [ 10071451177251346351593122552258400731070307792115572537969044314339076126231,
+    2894161621123416739138844080004799398680035544501805450971689609134516348045 ];
+
+    uint256 constant curveOrderBitLength = 254;
+    uint256 constant curveOrderBitShift = 255 - curveOrderBitLength;
```
jot2re, line 54: Due to my original mistake this should actually be 256- curveOrderBitLength

src/main/java/com/alphawallet/attestation/demo/SmartContract.java
```diff
@@ -2,38 +2,47 @@
...
 
-  //test contract
-  public boolean testEncoding(ProofOfExponent exp)
+  public boolean verifyEqualityProof(byte[] com1, byte[] com2, ProofOfExponent pok) throws Exception
+  {
+    Function function = verifyEncoding(com1, com2, pok.getDerEncoding());
+    return callFunction(function);
+  }
+
+  public boolean testEncoding(FullProofOfExponent exp)
```
jot2re, line 32: This should be UsageProofOfExponent. The FullProofOfExponent contains the riddle point as well, which is not needed when proving equality of commitment in attestation and Ticket. I just tried this locally and that made the test work.

I realise now the names are not super good, so feel free to rename UsageProofOfExponent to AttestationProofOfExponent and FullProofOfExponent to EqualityProofOfExponent.

src/main/java/com/alphawallet/attestation/demo/Demo.java
```diff
@@ -227,10 +227,11 @@ private static void receiveCheque(Path pathUserKey, Path chequeSecretDir,
     }
     // TODO how should this actually be?
     SmartContract sc = new SmartContract();
+    /*FullProofOfExponent newPok = new FullProofOfExponent(redeem.getDerEncoding());
```
jot2re, line 230: This should actually be UsageProofOfExponent. If that is used then things seem work as is.

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,633 @@
...
+    function mapTo256BitInteger(bytes memory input) public pure returns(uint256 res)
+    {
+        bytes32 idHash = keccak256(input);
+        res = uint256(idHash);
+    }
+
+    function mapToCurveMultiplier(bytes memory input) public pure returns(uint256 res)
+    {
+        bytes memory nextInput = input;
+        do
```
jot2re, line 268: I think it is better to avoid the loop here to prevent gas leakage in case someone has not constructed the proof correctly s.t. c<curveOrder. Instead the smart contract should just terminate

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,633 @@
...
+
+        return ecEquals(lhs, rhs);
+    }
+
+    function ecEquals(uint256[2] memory ecPoint1, uint256[2] memory ecPoint2) private pure returns(bool)
+    {
+        return (ecPoint1[0] == ecPoint2[0] && ecPoint1[1] == ecPoint2[1]);
+    }
+
+    function decodeAttestation(bytes memory attestation) public view returns(bool)
```
jot2re, line 156: I am a bit unsure about this function and the flow. Is this actually expected an Attestation or a usage request? The reason I am confused is because of the proof decoding below which should not be part of an Attestation, but only an attestation usage request, but in that case the riddle is not included now.

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
```diff
@@ -0,0 +1,633 @@
...
+        owner = msg.sender;
+        callCount = 0;
+    }
+
+    struct Length {
+        uint decodeIndex;
+        uint length;
+    }
+
+    struct Exponent {
```
jot2re, line 77: Maybe rename to FullProofOfExponent

jot2re (changes requested):
There are a couple of things I am confused about, maybe you can clarify these? I also had a suggestion to how to fix the test issue.
Finally I realise that I have made slight mistake when I rebased compressing-transaction with main since some part of the rejection sampling got missing from it. I am adding an update to compressing-transaction now. So please rebase to ensure we are synched

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
JamesSmartCell: This function is no longer used. I left it in for reference. Since this file hasn't yet got a git history it's not lost by leaving it in the original commit. It can be removed later.

I'll update it though.

src/main/java/com/alphawallet/attestation/demo/Demo.java
JamesSmartCell: Ok

src/test/java/com/alphawallet/attestation/demo/VerifyAttestation.sol
JamesSmartCell: It should be renamed as verifyAttestationRequestProof

colourful-land (changes requested):
Your test code doesn't work:

````$ gradle test

> Task :compileJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :compileTestJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :test

ProofOfKnowledgeTest > TestContract() FAILED
    org.opentest4j.AssertionFailedError at ProofOfKnowledgeTest.java:118

86 tests completed, 1 failed

> Task :test FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':test'.
> There were failing tests. See the report at: file:///home/a/IdeaProjects/attestation/build/reports/tests/test/index.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 6m 41s
3 actionable tasks: 3 executed
````

(note, I'll commit the change to move the test to IntegrationTest so when you want to rerun, if you pulled, you need to rerun with `gradle intTest`

**Comment by JamesSmartCell** [comment]
@jot2re I removed all the SmartContract calls from the ProofOfKnowledge test as they should be in the integration test. I believe they shouldn't be in there any longer.

Ideally for integration test: We create a docker instance with a Ganash eth node, deploy the latest contract then conduct the smart contract tests using that local node. This will be a lot faster than calling Kovan about 1100 times. This an be done later though.

File: src/main/java/com/alphawallet/attestation/demo/SmartContract.java [hunk 7]
```diff
@@ -29,7 +29,7 @@ public boolean verifyEqualityProof(byte[] com1, byte[] com2, ProofOfExponent pok
     return callFunction(function);
   }
 
-  public boolean testEncoding(FullProofOfExponent exp)
+  public boolean usageProofOfExponent(FullProofOfExponent exp)
   {
     Function function = checkEncoding(exp.getDerEncoding());
```
jot2re, line 34: The check encoding function seems to call the old function, instead of verifyAttestationRequestProofPayable. But in any case, it seems it is only called from Demo, which should still verifyEqualityProof

src/main/java/com/alphawallet/attestation/demo/Demo.java
```diff
@@ -227,11 +227,11 @@ private static void receiveCheque(Path pathUserKey, Path chequeSecretDir,
     }
     // TODO how should this actually be?
     SmartContract sc = new SmartContract();
-    /*FullProofOfExponent newPok = new FullProofOfExponent(redeem.getDerEncoding());
-    if (!sc.testEncoding(redeem.getPok())) {
+    FullProofOfExponent newPok = new FullProofOfExponent(redeem.getDerEncoding());
```
jot2re, line 230: I think we might have misunderstood each other. The redeem request only contains a UsageProofOfKnowledge, which is what should be given as argument to the smart contract along with the Ticket and Attestation commitments. 
What is happening now is that you are trying to construct a FullProofOfExponent from a UseDevconTicket object. This fails with an exception, but becasuse the Demo class actually handles all exceptions this is not seen by the JUnit test.
This is a mistake of mine; I didn't verify that the test would fail when an internal method fails. I have committed a small change to Demo on this branch to ensure that the test fails when something goes wrong.

jot2re (changes requested):
Most of the changes are handled. However there is still a little misunderstanding between us that should be handled. In fact, because of an old mistake of mine, a test failed to catch an error that arrives from this misunderstanding.

Finally, two minor things from the previous review looks like they are still missing:
1. Renaming of the Exponent structure in the VerifyAttesation.sol
2. mapToCurveMultiplier in  VerifyAttesation.sol still seems to allow for too much gas usage in case a proof has not been constructed correctly. I.e. it still has the loop where it tries to compute the challenge c if it is not properly sampled. But this is not critical as discussed in last review.

src/main/java/com/alphawallet/attestation/demo/Demo.java
JamesSmartCell: OK, I didn't check what was going on there; since the test doesn't work from gradle the way it is set up.

src/main/java/com/alphawallet/attestation/demo/SmartContract.java
JamesSmartCell: Ah, I accidentally git overrode the fix to mapToCurveMultiplier which had the loop removed!

src/main/java/com/alphawallet/attestation/demo/Demo.java
JamesSmartCell: Fixed. I actually was editing the wrong contract, confusion as the file positions and names keep jumping around all over the repo.

jot2re (changes requested):
Looks good James, but I still think there are a couple of missing commits to the Demo.java and smartcontract.java in relation to the last review and your comments on the last review.

colourful-land:
> @jot2re I removed all the SmartContract calls from the ProofOfKnowledge test as they should be in the integration test. I believe they shouldn't be in there any longer.
> 
> Ideally for integration test: We create a docker instance with a Ganash eth node, deploy the latest contract then conduct the smart contract tests using that local node. This will be a lot faster than calling Kovan about 1100 times. This an be done later though.

yes but make an issue instead of just mentioning it here.

colourful-land:
Set aside the renaming needed in https://github.com/TokenScript/attestation/issues/107 for @JamesSmartCell so this one can close.

PR #93, opened 2021-01-07
Polyfill bigint

nicktaras:


colourful-land:
I believe this is a draft PR because Nick mentioned he would try to not include polyfill here (but in Negotiator library instead, or the website repo).

micwallace:
The usage of polyfills & node libraries is a bit of an issue. We can't expect websters to have to include these polyfills in their bundler configuration. I think we should start an effort to remove usage of node built-ins where possible and add any polyfills here for older browser support.

micwallace:
No longer required as all modern browsers now support bigint

PR #97, opened 2021-01-08
Remove 67 lines of code by avoiding the 90s file io API

colourful-land:
By the way I think in the test cases we don't worry overwriting test output files - in fact avoiding overriting makes test ambiguous, since `gradle testClean && gradle test` is expected to write new contents instead of shying away from overwriting.

### DIFF
--- a/src/main/java/com/alphawallet/attestation/core/DERUtility.java
+++ b/src/main/java/com/alphawallet/attestation/core/DERUtility.java
@@ -17,6 +17,7 @@ import org.bouncycastle.util.encoders.Base64Encoder;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


src/test/java/com/alphawallet/attestation/cheque/TestRedeemCheque.java
```diff
@@ -71,25 +71,25 @@ public void testSunshine() {
     try {
       PublicKey pk;
       System.out.println("Signed attestation:");
-      System.out.println(DERUtility.printDER(attestedCheque.getAtt().getDerEncoding(), "SIGNEABLE"));
+      System.out.println(DERUtility.toPEM(attestedCheque.getAtt().getDerEncoding(), "SIGNEABLE"));
```
jot2re, line 74: The tests now only print out the object reference and not the actual PEM file in human readable base 64.

src/test/java/org/devcon/ticket/UseTicketTest.java
```diff
@@ -68,26 +68,26 @@ public void testSunshine() {
     try {
       PublicKey pk;
       System.out.println("Signed attestation:");
-      System.out.println(DERUtility.printDER(attestedTicket.getAtt().getDerEncoding(), "SIGNABLE"));
+      System.out.println(DERUtility.toPEM(attestedTicket.getAtt().getDerEncoding(), "SIGNABLE"));
```
jot2re, line 71: The tests now only print out the object reference and not the actual PEM file in human readable base 64.

src/main/java/com/alphawallet/attestation/core/DERUtility.java
```diff
@@ -104,30 +104,19 @@ public static BigInteger decodeSecret(byte[] secretBytes) {
...
       coder.encode(input, 0, input.length, outstream);
       byte[] encodedCert = outstream.toByteArray();
-      StringBuilder builder = new StringBuilder();
-      builder.append("-----BEGIN " + type + "-----\n");
-      addBytes(builder, encodedCert);
-      builder.append("-----END " + type + "-----");
-      return builder.toString();
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream( );
+      outputStream.write(("-----BEGIN " + type + "-----\n").getBytes(StandardCharsets.UTF_8));
+      outputStream.write(encodedCert);
```
jot2re, line 115: No new-line information gets encoded which is needed for valid PEM files the results in the command line demo not working anymore.

jot2re (changes requested):
It is nice get get some lines of code removed by taking advance of newer API methods, but unfortunately the logic is changed a bit which result in issues with the command line Demo.

src/test/java/com/alphawallet/attestation/cheque/TestRedeemCheque.java
colourful-land: Thanks fixed.

src/main/java/com/alphawallet/attestation/core/DERUtility.java
colourful-land: new line added thanks.

jot2re (changes requested):
It looks good and works now. However I still think that the linebreaks are strictly needed if we want to be PEM compliant. See https://tools.ietf.org/html/rfc1421#section-4.3.2.4 which says that the lines containing the encoding must be 64 characters long

colourful-land:
> It looks good and works now. However I still think that the linebreaks are strictly needed if we want to be PEM compliant. See https://tools.ietf.org/html/rfc1421#section-4.3.2.4 which says that the lines containing the encoding must be 64 characters long

Oh, fixed plus findings:

1. bouncy castle didn't provide a line-wrapping encoder - had to use the one shipped with Java, strange because they should be more compatible with that rfc than Java's built in encoder.
2. the code at `main` branch actually wraps at 65th character instead of 64.

jot2re (approved):
Looks great now, thank!

PR #101, opened 2021-01-22
Webauth

### LINKED ISSUE #99: create the web authentication backend at org.tokenscript.auth
Verify SignedDevconTicket (against known pubkey)
Verify MyAttestation (against known pubkey)
Verify the zk proof
Verify the challenge is in the message
Verify the signature is attested by MyAttestation.

The “Message” is signed by wallet (e.g. metamask)
therefore have a prefix. It might be a json array with
Base64 encoded UseDevconTicket:
{“useDevconTicket”: “Ze8zE/==”, “challenge”, “809843…”}

jot2re:
I have implemented a class for verifying useDevconTicket requests.
Basically the class takes as input a JWT containing a useDevconTicket along with a timestamp and a domain that has been signed with a user's Ethereum key (with the "Ethereum signed message" prefix).
It verifies useDevconTicket and the time stamp and signature. 
I also created a simple JWT issuer that constructs such a request. @oleggrib this JWTIssuer eventually needs to be implemented in Javascript. I found several posts online outlining how to do JWTs with Ethereum keys so it should not be too difficult. 

This closes #99.


jot2re:
I have implemented EIP712 for metaMask compatible web authentication.
As part of this I have refactored AttestationCrypto and SignatureUtility by moving signature related things into SignatureUtility and fully implemented Ethereum style signatures (s value less than < n/2, and computation of recovery value based EIP155). 
In order to support EIP712 I have imported updated code from AlphaWallet.

PR #104, opened 2021-01-25
Implemented capability issuer and validator using JWT

jot2re:
This allows Devcon to issue tokens to websites that can be verified in a devcon.ticket iframe.


src/main/java/com/alphawallet/attestation/core/SignatureUtility.java
```diff
@@ -56,6 +67,45 @@ public static AsymmetricKeyParameter restoreKeyFromSPKI(byte[] input) throws IOE
         return PublicKeyFactory.createKey(spki);
     }
 
+    public static PrivateKey PrivateBCKeyToJavaKey(AsymmetricKeyParameter bcKey) {
```
aimiebell, line 70: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

jot2re: Handled

jot2re:
@colourful-land would you be able to take a look at this PR soon? PS since webauth has already been merged, this should now be merged directly into main.

colourful-land:
Need to merge to main

PR #105, opened 2021-01-25
Ensured that the Ethereum signing method is used all places where a u…

jot2re:
…ser would sign using their Ethereum key.

This is to ensure consistency with MetaMask when things will integrate with Hu and Oleg's code. This closes #105

### DIFF
--- a/src/main/java/com/alphawallet/attestation/IdentifierAttestation.java
+++ b/src/main/java/com/alphawallet/attestation/IdentifierAttestation.java
@@ -10,6 +10,7 @@ import org.bouncycastle.asn1.ASN1EncodableVector;
 import org.bouncycastle.asn1.ASN1ObjectIdentifier;
 import org.bouncycastle.asn1.DEROctetString;
 import org.bouncycastle.asn1.DERSequence;
+import org.bouncycastle.asn1.x509.AlgorithmIdentifier;
 import org.bouncycastle.asn1.x509.SubjectPublicKeyInfo;
 import org.bouncycastle.crypto.params.AsymmetricKeyParameter;
 import org.bouncycastle.crypto.util.PublicKeyFactory;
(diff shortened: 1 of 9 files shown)


src/test/java/com/alphawallet/attestation/IdentifierAttestationTest.java
```diff
@@ -64,7 +64,7 @@ public void testCannotSet() {
       // Expected
     }
     try {
-      initial.setSignature("1.0.12.151");
+      initial.setSigningAlgorithm("1.0.12.151");
```
colourful-land, line 67: What is this?? This OID look way too short

src/main/java/com/alphawallet/attestation/IdentifierAttestation.java
```diff
@@ -31,7 +30,7 @@ public IdentifierAttestation(String identity, AttestationType type, AsymmetricKe
     super();
     super.setVersion(18); // Our initial version
     super.setSubject("CN=" + AttestationCrypto.addressFromKey(key));
-    super.setSignature(AttestationCrypto.OID_SIGNATURE_ALG);
+    super.setSigningAlgorithm(AttestationCrypto.OID_SIGNATURE_ALG);
```
colourful-land, line 33: Wait, OID_SIGNATURE_ALG is currently set to "1.2.840.10045.2.1" which is `ecPublicKey` according to this: https://oidref.com/1.2.840.10045.2.1 - which should be a component of PublicKeyInfo instead of attestation's signing algorithm?

jot2re: I have not made any change to the OIDs for a very long time. Initially when we tried to make everything work with X509 (in particular with Java's own X509 code) I think this was one of the only reasonable parameters we could set that actually worked. But as you say it is not completely correct and the reason is that there is no OID for what we actually do. 
However since we are actually not currently verifying things in a smart contract there is no reason why the attestation must be signed with Keccak256 and secp256k1. 
I can create an issue investigating this further and perhaps add support for standard signatures if you think it is relevant?

PR #111, opened 2021-02-04
Blockchain integration bug

### LINKED ISSUE #109: Failure in integrating new PoKs with the smart contract
With the recent merge of the updates to smart contract code and optimisations, it turns out that the smart contract integration does not work completely with the java implementation any more (or that the java code has not been updated to use the new smart contract).
This can be seen when running the tests in TestDemo in the TestDemo.java file and SmartContractTest.java.
It turns out this is most likely an issue with the updated format of the ProofOfExponent.
I have tried to update the java fixtures (integrating with my code) to reflect the updated structure, but I think there are still some integration changes that should be updated.

I have implemented the change in the branch blockchain-integration-bug which should be used for finishing this issue.

JamesSmartCell:
- Update contract with correct curve params
- Update proof forming method with the rejection sampling mods.
- Update contract address.

closes #109

### DIFF
--- a/src/intTest/java/com/alphawallet/attestation/SmartContractTest.java
+++ b/src/intTest/java/com/alphawallet/attestation/SmartContractTest.java
@@ -113,19 +113,23 @@ public class SmartContractTest {
             assertFalse(negativeCheck);
         }
 
-        // FullProofOfExponent attestation request proof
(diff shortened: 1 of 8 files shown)


jot2re (approved):
Thanks a lot! Looks good!

Issue #112, opened 2021-02-22
Implement Eip712 signing for all cases where the user needs to sign


jot2re:
So far the reference implementation only uses EIP712 for web authentication. To make the whole flow user-friendly and pretty this message format should be implemented all places where the user needs to sign. This includes:
1. Attestation request.
2. Attested object.

colourful-land:
User-friendliness is not the only thing we should watch for. Today's transaction costs $80 USD easily. Gas....

When authenticating for the web, or doing capability attestation, there is no gas consideration so we should maximise user experience

jot2re:
Auch! That is a huge price increase. I still think item 1 is valid, since this is never going to be posted to the blockchain as you also point out. In regards to item 2  it is currently only used unsigned in the webauthentication flow (since it is just a component of what is signed). So with you comment in mind I think we should not touch the signing of this (at least for now).
I am also realising another issue with using EIP712 and DER together. The DER encoded content will not make sense in the EIP712 interface and thus the relevant, human-readable aspects of what is DER encoded should then be included in the EIP712 message which will make the message even bigger...

jot2re:
Closed (so far) by PR #113

PR #113, opened 2021-02-25
Eip712 attestation request

jot2re:
I have refactored the code for Eip712 used for web authentication to make it easier to use for other data structures.
Then I have changed the DER structure of AttestationRequest to only include non-human readable info and made a new class for an EIP712AttestationRequest that will work well with MetaMask. This allows verification and construction of EIP712 Attestation Requests in a way where all the cryptography and validations are still in place and where the user sees something sensible to sign.
This closes issue #112


colourful-land (approved):
Thanks very much

PR #120, opened 2021-03-25
Smart Contract Attestation Decoding

JamesSmartCell:
This is a first draft of decoding attestation to check Subject and Attestor addresses.

Not currently supported:

- Checking of timestamp validation


jot2re (approved):
Looks good for decoding an x509 type identity attestation. Note that no further validity checks on the content is currently being done in this solidity code.

PR #121, opened 2021-03-31
Add Public Identifier Attestation

JamesSmartCell:
First draft for a public attestation which is confirming a cryptographic link between an ethereum address and a public identifier eg a Twitter ID.


src/main/java/com/alphawallet/attestation/IdentifierAttestation.java
```diff
@@ -61,6 +61,24 @@ public IdentifierAttestation(byte[] commitment, AsymmetricKeyParameter key)  {
...
+      SubjectPublicKeyInfo spki = SubjectPublicKeyInfoFactory.createSubjectPublicKeyInfo(key);
+      super.setSubjectPublicKeyInfo(spki);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+
+    //Add the identifier in the form 'TW=@twittertest' or 'EMAIL=test@test.org'
+    ASN1EncodableVector dataObject = new ASN1EncodableVector();
+    dataObject.add(new DEROctetString((type + "=" + identifier).getBytes()));
+    super.setDataObject(new DERSequence(dataObject));
```
jot2re, line 79: This should work just fine, but I think maybe it is better to also store this as an extension in a similar manner to the commitment, because this allows us to follow the x509 standard more closely. But for all our internal code this should work just fine

jot2re (approved):
This should work just fine, but I think maybe it is better to also store this as an extension in a similar manner to the commitment, because this allows us to follow the x509 standard more closely. But for all our internal code this should work just fine

colourful-land:
@jot2re there is some practical trouble following x.509. How do you think we should address it?

The problem is with the subject. According to x.509 subject look like these 2 examples:

````
cn=aw.app
mail=james.sangalli@alphawallet.com
````

But since @JamesSmartCell is working on "public attestation" the subject should be a public string, such as the first example, and not a string with data to hide, such as the second example.

So ideally it should be:

````
twitter=zhangweiwu
````

The problem is that there is no ASN.1 attribute for twitter (no OID). I'm sending an email to other opensource projects to see how they handles it.

**Comment by jot2re** [comment]
> @jot2re there is some practical trouble following x.509. How do you think we should address it?
> 
> The problem is with the subject. According to x.509 subject look like these 2 examples:
> 
> ```
> cn=aw.app
> mail=james.sangalli@alphawallet.com
> ```
> 
> But since @JamesSmartCell is working on "public attestation" the subject should be a public string, such as the first example, and not a string with data to hide, such as the second example.
> 
> So ideally it should be:
> 
> ```
> twitter=zhangweiwu
> ```
> 
> The problem is that there is no ASN.1 attribute for twitter (no OID). I'm sending an email to other opensource projects to see how they handles it.

It is a very good question! So far we have simply just abused the extensions field to add any custom information that can still be interpreted by any parser of x509. An alternative could be to use (Subject Alternative Names as part of the Common Name)[https://tools.ietf.org/html/rfc5280#section-4.2.1.6]
However, it again has the issue of not exactly allowing the ``twitter=zhangweiwu`` approach that we want.

**Comment by JamesSmartCell** [comment]
Should the attestation period be included? Since the attestation will be linked to the wrapped NFT and the main focus of this style is to be compliant with existing infrastructure; it makes sense for the attestation to have no validity period:

eg. Bob asks Charlie Midas to wrap his NFT.

Bobs keeps his NFT for 20 years and presents it as a gift to Alice. When she runs the attestation through the cryptography infrastructure, it should return valid without needing any custom interpretation.

If the attestation is time limited, as the single-use attestations created by the Devcon are; then this won't work.

Remove the time limits? Or just remove the expiry?

**Comment by jot2re** [comment]
> Should the attestation period be included? Since the attestation will be linked to the wrapped NFT and the main focus of this style is to be compliant with existing infrastructure; it makes sense for the attestation to have no validity period:
> 
> eg. Bob asks Charlie Midas to wrap his NFT.
> 
> Bobs keeps his NFT for 20 years and presents it as a gift to Alice. When she runs the attestation through the cryptography infrastructure, it should return valid without needing any custom interpretation.
> 
> If the attestation is time limited, as the single-use attestations created by the Devcon are; then this won't work.
> 
> Remove the time limits? Or just remove the expiry?

RFC-5280 (specifying X509) mentions how you can indicate an un-limited lifetime. So I suggest we simply do that :)

``To indicate that a certificate has no well-defined expiration date,
   the notAfter SHOULD be assigned the GeneralizedTime value of
   99991231235959Z.``

**Comment by JamesSmartCell** [comment]
> To indicate that a certificate has no well-defined expiration date, the notAfter SHOULD be assigned the GeneralizedTime value of 99991231235959Z.

Neat, I'll update the attestation with that.

**Comment by colourful-land** [comment]
@JamesSmartCell I have talked with @jot2re and here is the thinking.

1. We will use lableed URI for the OID in the subject. @jot2re is going to update it.
2. We will stay compatible with x.509 for now, which means one more round of kaccek from Ethereum subjectPublicKey to Ethereum address (cost 30~42 gas).

**Comment by JamesSmartCell** [comment]
So Transmogrify object (the signed wrapper of original NFT(s), Public Identifier Attestation and Autograph SHA256) is going to be x509 or EIP712 (SignTypedMessage) ? There's advantages and disadvantages in using both these formats.

**Comment by jot2re** [comment]
> So Transmogrify object (the signed wrapper of original NFT(s), Public Identifier Attestation and Autograph SHA256) is going to be x509 or EIP712 (SignTypedMessage) ? There's advantages and disadvantages in using either of these formats. Also can you clarify what 'labelled URI' is in this case?

Since I assume it is going to be used on the blockchain, I imagined it would be a custom DER object containing the data you mentioned above.

**Comment by colourful-land** [comment]
> So Transmogrify object (the signed wrapper of original NFT(s), Public Identifier Attestation and Autograph SHA256) is going to be x509 or EIP712 (SignTypedMessage) ? There's advantages and disadvantages in using either of these formats. Also can you clarify what 'labelled URI' is in this case?

Perhaps not - I don't feel TransObject should be in x.509. x.509 is mostly about identity. But both will be in DER. 

x.509 and DER is not the same thing. x.509 uses DER. Your current public identifier attestation is DER but not x.509.

**Comment by JamesSmartCell** [comment]
OK then, I'll just structure the token data, public identifier attestation, autograph hash and wrapping signature as a DER. That should contain everything we need and be portable enough.

**Comment by colourful-land** [comment]
> OK then, I'll just structure the token data, public identifier attestation, autograph hash and wrapping signature as a DER. That should contain everything we need and be portable enough.

I think so. For identifier attestation, being compatible with standards potentially allow reuse, such as twitter account holder signs a message that proves he is the accoun holder using such x.509. But such reusability isn't seen in other attesatations, such as transObject which you are making.

File: src/main/java/com/alphawallet/attestation/NFTAttestation.java [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+import org.bouncycastle.asn1.ASN1EncodableVector;
+import org.bouncycastle.asn1.ASN1Primitive;
+import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.DERSequence;
+
+public class NFTAttestation implements ASNEncodable, Validateable {
+    private final SignedAttestation att;
+    private final ASN1Sequence token;
+
+    public NFTAttestation(SignedAttestation att, ERC721Token nftToken) {
```
jot2re, line 15: I think we still need the optional autograph as a digest value

src/main/java/com/alphawallet/ethereum/ERC721Token.java
```diff
@@ -0,0 +1,53 @@
...
+    {
+        this.address = address;
+        this.tokenId = tokenId;
+    }
+
+    public ASN1EncodableVector getTokenVector()
+    {
+        ASN1EncodableVector data = new ASN1EncodableVector();
+        data.add(new DEROctetString(Numeric.hexStringToByteArray(address)));
+        data.add(new ASN1Integer(tokenId));
```
jot2re, line 39: Why is this encoded as an integer and not just a 256 bit string? (It can of course be both but I am asking in relation to the asn 1 definitions, whether we should do one or the other :) )

jot2re (commented):
Looks good. A few notes and questions and still a few things to implement. 
One final thing not mentioned is that we need some more tests once the classes are stable

src/main/java/com/alphawallet/ethereum/ERC721Token.java
JamesSmartCell: Because it's much faster to decode in Solidity.

If it's encoded as a string, we need to convert it back from ASCII to hex then to address. I have written this code in assembler as the gas cost was quite high for compiled Solidity. Previously we encoded the address as an ASCII string - so a pair of ASCII character becomes a nibble; two pairs a byte.

If it's an integer then it's a fraction of the cost, and can be done by pulling out the bytes directly, then manipulating the bytes to fit into an address variable, which is very cheap and clean.

src/main/java/com/alphawallet/attestation/NFTAttestation.java
JamesSmartCell: What is the Autograph in this case? Previously we were going to include the digest of the .png of the actual autograph but now it's a procedurally added Twitter name.

jot2re: On Friday @colourful-land mentioned that it would be a hash digest of the transmogrified  object.

src/main/java/com/alphawallet/ethereum/ERC721Token.java
jot2re: Okay, but in that case shouldn't the address also be an integer? 
Maybe we are misunderstanding each other, I don't mean an ASCII string. I mean an OctetString, just like you encode the  address here. The reason I am asking is because this allows us to specify a specific amount of bytes in ASN1, whereas an ASN1 Integer is unbounded.

JamesSmartCell: > Okay, but in that case shouldn't the address also be an integer?
> Maybe we are misunderstanding each other, I don't mean an ASCII string. I mean an OctetString, just like you encode the address here. The reason I am asking is because this allows us to specify a specific amount of bytes in ASN1, whereas an ASN1 Integer is unbounded.

Ah I see. OK let's change it to OctetString. Address could be an integer and that would be ok too - as bytes it's equally as easy to convert to an address.

JamesSmartCell:
@jot2re I have updated two points:

- refined the method to extract the SPKI data from the public key to use standard functions.
- updated ERC721 tokenID to use OctetString.
- 
Can you verify it's all good and I'll finally push this.

src/main/java/com/alphawallet/attestation/Attestation.java
```diff
@@ -309,6 +309,17 @@ public boolean checkValidity() {
     return attEncoded;
   }
 
+  public byte[] getKeyDER()
```
jot2re, line 312: It looks like this method is never used or tested?

src/main/java/com/alphawallet/attestation/NFTAttestation.java
```diff
@@ -1,22 +1,33 @@
...
+    {
         this.att = att;
         this.token = new DERSequence(nftToken.getTokenVector());
     }
 
+    public NFTAttestation(byte[] derEncoding, AsymmetricKeyParameter signingPublicKey) throws IOException {
+        ASN1InputStream input = new ASN1InputStream(derEncoding);
+        ASN1Sequence asn1 = ASN1Sequence.getInstance(input.readObject());
+
+        ASN1Sequence attestationEnc = ASN1Sequence.getInstance(asn1.getObjectAt(0)); //root attestation, should be signed att
```
jot2re, line 26: This does not completely match the latest asn1 format [we have](https://github.com/TokenScript/attestation/blob/main/data-modules/output/asn/NFTAttestation.asn). However if what is in the code is what we want the it is super easy for me to change the asn1 description. But we should be sure this is the format we actually want

src/main/java/com/alphawallet/attestation/NFTAttestation.java
```diff
@@ -29,7 +40,15 @@ public NFTAttestation(SignedAttestation att, ERC721Token nftToken) {
         }
     }
 
+    public byte[] getPreHash() {
+        return SignatureUtility.convertToPersonalEthMessage(getDerEncoding());
```
jot2re, line 44: This is a bit inconsistent with what is done in Attestation and the other classes. That is, in Attestation the raw DER data is returned and then it is the responsibility of the signing method to add the Personal prefix if needed. This is for example done in AttestedObject. But maybe the prehash in Attestation should return the DER data with the Personal prefix added?

src/test/java/com/alphawallet/attestation/PublicAttestationTest.java
```diff
@@ -30,8 +32,30 @@ public static void setupKeys() throws Exception {
...
+
+        //generate NFTAttestation from the NFTAttestation bytes
+        NFTAttestation nftAttestation2 = new NFTAttestation(nftAtt.getDerEncoding(), attestorKeys.getPublic());
+
+        //check recovered signed attestation within the wrapping
+        assertTrue(nftAttestation2.verify());
+
+        //Generate SignedNFTAttestation using the reconstructed NFTAttestation and the extracted Ethereum signature
+        SignedNFTAttestation signedNFTAttestation2 = new SignedNFTAttestation(nftAttestation2, subjectKeys.getPublic(), sig);
+        assertTrue(signedNFTAttestation2.checkValidity());
```
jot2re, line 58: Should probably also have a call `assertTrue(signedNFTAttestation2.verify());`

src/test/java/com/alphawallet/attestation/PublicAttestationTest.java
```diff
@@ -48,7 +72,9 @@ public void testDecoding() throws Exception {
         assertTrue(SignatureUtility.verifyEthereumSignature(att.getPrehash(), signed.getSignature(), issuerKeys.getPublic()));
         assertArrayEquals(att.getPrehash(), signed.getUnsignedAttestation().getPrehash());
         byte[] signedEncoded = signed.getDerEncoding();
-        SignedAttestation newSigned = new SignedAttestation(signedEncoded, issuerKeys.getPublic());
-        assertArrayEquals(signed.getDerEncoding(), newSigned.getDerEncoding());
+
+        //Fails?
+        //SignedAttestation newSigned = new SignedAttestation(signedEncoded, issuerKeys.getPublic());
```
jot2re, line 77: `HelperTest.makeMaximalAtt` uses current time to set how long it should be valid for. I think it would be better just to change `HelperTest.makeMaximalAtt` to be valid as long as possible to ensure the method is idempotent

jot2re (changes requested):
Looks good, but I do have some questions and comments in regarding to general consistency. One other thing is that it seems that currently 40 tests fails, but I think this might been resulted from some fundamental changes made on the main branch when it was merged in. So hopefully merging again will help this.
One final thing I only now noticed is that in SignedNFTAttestation.constructSignedAttestation the ASN1 spec we have does not include SignatureUtility.EC_PUBLIC_KEY_IDENTIFIER, since this is assumed implicit.

src/main/java/com/alphawallet/attestation/NFTAttestation.java
JamesSmartCell: Yes that's a good point. It's going to be a bit awkward at some end whichever way we do it; if you add the signpersonal prefix in the attestation, when you go to check it at the other end, a dapp developer might be concerned why they aren't checking via signpersonal, however the request was made via signpersonal. 

It was a quandary whether to do this or not - I thought it would make more sense for the attestation to appear to the dapp developer as being needed to be wrapped, since they're going to have to call signpersonalmessage instead of just 'sign'.

src/test/java/com/alphawallet/attestation/PublicAttestationTest.java
JamesSmartCell: ... it does have. Did you mean the original 'nftAttestation' SignedNFTAttestation? I can add that. Both are redundant however, because 'verify()' is called at the end by the constructor.

jot2re: I mean on SignedNFTAttestation. 
It is true that it is redundant since it is currently called during construction. However, tbh I am a bit unhappy abut my initial design choice of always doing so. As I don't think it should be expected by the constructor, and it makes negative testing much harder. Anyway, I am thinking that we might eventually want to remove this constructor check, so calling it here is more about future-proofing and making explicit what is expected of the object.

src/test/java/com/alphawallet/attestation/PublicAttestationTest.java
JamesSmartCell: I think I follow - is the reason it fails because the maximalAtt has a slightly different format?

src/test/java/com/alphawallet/attestation/PublicAttestationTest.java
```diff
@@ -55,6 +62,62 @@ public void testNFTAttestation() throws Exception
...
+        //check our return
+        assertEquals(atr.identity, attestationIdentifier);
+        assertEquals(atr.ownerAddress.toLowerCase(), SignatureUtility.addressFromKey(subjectKeys.getPublic()).toLowerCase());
+        assertTrue(atr.isValid);
+        ERC721TokenEth token = atr.ercToken[0];
+        assertEquals(token.address.toString().toLowerCase(), myNFT.address.toLowerCase());
+        assertEquals(myNFT.tokenId, token.tokenId.getValue());
+
+        //TODO: make a more comprehensive negative test, involving bad attestation subject address etc.
+        //TODO: can do this if we remove the 'isValid' check in the constructor
```
jot2re, line 116: I think it would definitely make things easier in regards to such negative tests

jot2re:
Looks good! I handled some tests issues and some weird renaming issues.
The only reaming issues/questions are the following:
1. Should Attestation.getKeyDER still be there since it is not tested or used?
2. Should NFTAttestation.java be updated to match NFTAttesation.asn or should NFTAttestation.java be used as reference?
3. TestAttestationUse.testSendAttestation still fails due to incorrect expected number of smart contracts. @JamesSmartCell can you take a look at this?

JamesSmartCell:
> Looks good! I handled some tests issues and some weird renaming issues.
> The only reaming issues/questions are the following:
> 
> 1. Should Attestation.getKeyDER still be there since it is not tested or used?
> 2. Should NFTAttestation.java be updated to match NFTAttesation.asn or should NFTAttestation.java be used as reference?
> 3. TestAttestationUse.testSendAttestation still fails due to incorrect expected number of smart contracts. @JamesSmartCell can you take a look at this?

Fixed points 1. and 3.

2. The difference is that it isn't an array, right? What's the best way to fix that?

jot2re:
> > Looks good! I handled some tests issues and some weird renaming issues.
> > The only reaming issues/questions are the following:
> > 
> > 1. Should Attestation.getKeyDER still be there since it is not tested or used?
> > 2. Should NFTAttestation.java be updated to match NFTAttesation.asn or should NFTAttestation.java be used as reference?
> > 3. TestAttestationUse.testSendAttestation still fails due to incorrect expected number of smart contracts. @JamesSmartCell can you take a look at this?
> 
> Fixed points 1. and 3.
> 
> 1. The difference is that it isn't an array, right? What's the best way to fix that?

Great, thanks! In regards to 2, the difference is both that it is supposed to take an array of tokens and that it should take an optional digest. 
In regards to the array part it should be sufficient to simply encode this as a SEQUENCE of tokens.

JamesSmartCell:
> Great, thanks! In regards to 2, the difference is both that it is supposed to take an array of tokens and that it should take an optional digest.
> In regards to the array part it should be sufficient to simply encode this as a SEQUENCE of tokens.

This is ```DERSequence``` you're referring to? Do you know if it's possible to determine the array size before you start reading the contents? I can check it but I have a large task stack.

jot2re:
Yeah, 

> > Great, thanks! In regards to 2, the difference is both that it is supposed to take an array of tokens and that it should take an optional digest.
> > In regards to the array part it should be sufficient to simply encode this as a SEQUENCE of tokens.
> 
> This is `DERSequence` you're referring to? Do you know if it's possible to determine the array size before you start reading the contents? I can check it but I have a large task stack.

Yes, if you look at page 10 [here ](https://www.itu.int/ITU-T/studygroups/com17/languages/X.690-0207.pdf) you see that after 0x30 the length of the sequence is encoded.

colourful-land:
> This can't merge yet, there are problems.

Solved.

PR #124, opened 2021-04-15
Address and default time fixes

jot2re:
Made a fix to nonce handling to always use fully upper-case addresses. Furthermore I changed the default validity time for attestations to be 1 year and did a bit of refactoring


oleggrib (approved):
All Ok here, I have added same changes to the JS code

PR #138, opened 2021-06-04
Identity to identifier

jot2re:
I have refactored both the java, asn and type script code. The only thing I haven't touched is the bundle.js code. So that might need to be rebundled @oleggrib


oleggrib (approved):

colourful-land:
you need to remove the bundle.js and make sure it is regenerated with npm build, and please help solving the confilcits

colourful-land:
@samuelhbne can you help solve conflict? Plate too full for me

PR #142, opened 2021-06-21
Fully verify attestation w usage

jot2re:
Fixes a few cosmetic issues in relation to the command line utility and ensured that full verification is done when using the approach where a user constructs a combined attestation and usage request in a single go.


colourful-land (approved):
Thanks

PR #143, opened 2021-06-21
Added magic-link utility to cmd

jot2re:


colourful-land:
Instead of using

````
     System.out.println(URLUtility.encodeData(signature));
````

Which, I believe, will cause `/` to be translated to `%2F`, you should follow the example in https://github.com/TokenScript/attestation/blob/1086a4a6b9911c7bd9219aaf66493b75bd4901a2/src/main/java/org/devcon/ticket/TicketIssuer.java#L48

colourful-land (changes requested):
Instead of using

````
     System.out.println(URLUtility.encodeData(signature));
````

Which, I believe, will cause `/` to be translated to `%2F`, you should follow the example in https://github.com/TokenScript/attestation/blob/1086a4a6b9911c7bd9219aaf66493b75bd4901a2/src/main/java/org/devcon/ticket/TicketIssuer.java#L48

jot2re:
Thanks for the feedback. I have changed the code accordingly

PR #145, opened 2021-06-23
Use External Signature with Ticket Attestation

JamesSmartCell:
- Adding ability to add signature to AttestedObject externally; ie fom a user signed system.
- Add tests for external signature.
- Add contract and contract test.


jot2re (approved):
Looks good! I guess since we have a lot of constructors now, we could refactor a bit to remove code-duplication in constructors. But it seems fine the way things are.
I added some exception wrapping, to make sure logging also happens in case something goes wrong.

PR #148, opened 2021-06-30
Unpredictible number reference in java

jot2re:


colourful-land:
I'm asking @fyang1024 to work directly on the same branch his JavaScript implementation, so if changes needed before merge they can be applied to both JS and Java version.

colourful-land (changes requested):
NFTOFFER.asn doesn't belong to this branch, kindly remove it and consolidate (git-rebase) if you feel the need.

jot2re:
Sorry about the extra file. I must accidentally have based this branch on another branch I was working on. The file is removed now.

src/main/java/org/devcon/ticket/UnpredictableNumberTool.java
```diff
@@ -0,0 +1,81 @@
...
+  public UnpredictableNumberBundle getUnpredictableNumberBundle() {
+    long expiration = Clock.systemUTC().millis() + validityInMs;
+    return new UnpredictableNumberBundle(getUnpredictableNumber(expiration), domain, expiration);
+  }
+
+  private String getUnpredictableNumber(long expirationInMs) {
+    hmac.reset();
+    hmac.update(longToBytes(expirationInMs), 0, Long.BYTES);
+    hmac.update(domain.getBytes(StandardCharsets.UTF_8), 0, domain.getBytes(StandardCharsets.UTF_8).length);
+    byte[] digest = new byte[256 / 8];
```
oleggrib, line 57: Can you  generate random number , add it to hmac and prepend that UN with that number?

Like that logic:
expiry: now + 3600
UN1: rand()
UN2: hmac(concat(UN1, expiry), bogotá-secret)
UN: concat(UN1, UN2)

jot2re: I have added randomness (and kept the domain name).

jot2re:
@colourful-land please check if it is ok now.

oleggrib (approved):

PR #149, opened 2021-07-01
compare negotiator

nicktaras:


colourful-land:
This is a huge change, not knowing the context should we assign @oleggrib to review this?

Issue #134, opened 2021-06-03
Add email id to magic link generating Java function


AW-STJ:
Append the attestation link generated with users email, using
&email=

PR #150, opened 2021-07-01
really fixing #134 -

colourful-land:
I didn't understand what #137 was to do, except that it surely had no impact on the Link Generator. This is perhaps because the link generator's document isn't in github but in Google Drive, so @jot2re lost track of where the code should be modified. Anyway this one really solves #134 where #137 didn't.


jot2re:
Sorry, for the misunderstanding. This looks good.

PR #166, opened 2021-07-20
adding component map for developers for source code management

colourful-land:
You may want to change wording a bit before merging thanks.


nicktaras:
Sorry for the delay. Looks great, thanks for helping to run through this diagram with me also.

Issue #153, opened 2021-07-02
remove the signing of useTicket 


colourful-land:
I think during the devcon demo I noticed useTicket being signed. if it is true, it should be skipped. both the smart contract and the front end should allow useTicke to be used without signed.

on another note, useTicket should be different for each transaction bounce, but that can be done in a separate issue.

oleggrib:
currently useTicket just combined ticket+attestation+proof without signature and it always same while attestaion not expire
useEthKey contains signature of the challenge

jot2re:
I agree with @oleggrib regarding this issue. But even though adding the signature ensures that the useTicket is different each time, it is still possible to reuse an old useTicket as it remains valid.  However as we have briefly discussed a bit previously simply adding an unpredictable to the proof can ensure that it is unique each time and that it can be linked to a specific use context. The java code is already prepared to take such a parameter and thus can be easily adapted

PR #168, opened 2021-07-27
Remove signatures from UseTicket

jot2re:
Handles #153

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Which will create a jar file that you can run in the commandline
 
     build/libs/attestation-all.jar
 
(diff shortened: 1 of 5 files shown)


jot2re:
In relation to removing the need for a signature on the UseTicket object I assume this is also the case for UseCheque objects (from our previous project). Since a UseCheque object would need to be posted to the blockchain to be cashed and hence it would implicitly contain a signature by the user who is trying to cash it

jot2re:
@JamesSmartCell I have removed the signature form UseTicket object. This affect one of your tests interacting with the blockchain. The test itself currently works when run locally but I imagine some Solidity code change is required. It is `UseTicketTest.testSmartContractDecode` and I have tagged you.

PR #173, opened 2021-08-19
Fix statuslog errors

jot2re:
This fixes issue #172 
It fixes and issue with the way the shadow jar was constructed using log4j data.
See https://stackoverflow.com/questions/48033792/log4j2-error-statuslogger-unrecognized-conversion-specifier
The fix furthermore include explicitly validating the Ticket constructed to be sure that if the magic link gets printed, it will be valid

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,6 +92,7 @@ task integrationTest(type: Test) {
 }
 
 shadowJar {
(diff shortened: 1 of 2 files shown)


colourful-land (approved):
Works for me.

PR #180, opened 2021-08-27
Updated gradle and code to ensure java 1.8 compatibility

jot2re:
Per request from Feng and Hu


colourful-land:
@foxgem what is the reason using Java 1.8? Last time Feng told me the deployment environment is Java 11.

colourful-land:
Closed to be reopened if we find ourselves having to target 1.8

PR #181, opened 2021-08-28
Updated gradle file for maven pushing

jot2re:
This only updates the gradle file to work properly with Maven central and to have the pushed library include all code including devcon and alchemynft


colourful-land (approved):
I'm okay providing all we have, but anything other than org.tokenscript.* shouldn't be used by projects importing this libraries. If you discover any project (ours) using non org.tokenscript.* code, it is a bug.

PR #182, opened 2021-08-30
Devcon updates and replaced console.log with logger to better control logging flow

oleggrib:


src/main/javascript/crypto/src/Authenticator.ts
```diff
@@ -117,7 +116,7 @@ export class Authenticator {
      *  - Since this token depends on identifier attestation, continue to open iframe to attestation.id who needs to provide the proof
      */
     getIdentifierAttestation() {
-        console.log('getIdentifierAttestation. create iframe with ' + this.attestationOrigin);
+        logger(3,'getIdentifierAttestation. create iframe with ' + this.attestationOrigin);
```
jot2re, line 119: Maybe it would make sense to replace the level with an enum reflecting the level? But I understand that it is a lot of changes to make for something cosmetic

src/main/javascript/crypto/src/libs/Attestation.ts
```diff
@@ -181,20 +182,28 @@ export class Attestation {
...
+            logger(1, "Some attest data missed");
             return false;
         }
         let currentTime = Date.now();
         let attNotBefore = this.getNotValidBefore();
         let attNotAfter = this.getNotValidAfter();
-        if ( attNotBefore && attNotAfter && !(currentTime >= attNotBefore && currentTime < attNotAfter)) {
-            console.log("Attestation is no longer valid");
+
+        if ( attNotAfter && !(currentTime < attNotAfter)) {
```
jot2re, line 192: Why is ALLOWED_ROUNDING not used here but used in line 197?

jot2re (changes requested):
Looks good. Just two comments I just want you to check before merging

jot2re (approved):

PR #185, opened 2021-09-30
Adding CoSignedIdentifierAttestation and tests.

JamesSmartCell:
Adding CoSignedIdentifierAttestation.

This attestation is the SignedIdentifierAttestation, signed again by an Ethereum Key owned by the same entity that owns the Identifier embedded in the SignedIdentifierAttestation.

The primary purpose of this is to be able to create a binding between a Social Media account and an Ethereum address.

### DIFF
--- a/src/main/java/org/twittertip/CoSignedIdentifierAttestation.java
+++ b/src/main/java/org/twittertip/CoSignedIdentifierAttestation.java
@@ -1,0 +1,81 @@
+package org.twittertip;
+
+import org.bouncycastle.asn1.ASN1EncodableVector;
+import org.bouncycastle.asn1.ASN1Primitive;
+import org.bouncycastle.asn1.DERBitString;
+import org.bouncycastle.asn1.DERSequence;
+import org.bouncycastle.crypto.AsymmetricCipherKeyPair;
+import org.bouncycastle.crypto.params.AsymmetricKeyParameter;
+import org.tokenscript.attestation.SignedIdentifierAttestation;
+import org.tokenscript.attestation.core.ASNEncodable;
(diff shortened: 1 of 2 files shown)


src/main/java/org/twittertip/CoSignedIdentifierAttestation.java
```diff
@@ -0,0 +1,81 @@
...
+        this.att = att;
+        this.signature = signature;
+        this.attestationVerificationKey = subjectPublicKey;
+        if (!verify()) {
+            throw new IllegalArgumentException("The signature is not valid");
+        }
+    }
+
+    public SignedIdentifierAttestation getUnsignedAttestation() {
+        return att;
```
jot2re, line 42: I think this method signature is a bit misleading, since the attestation returned is actually signed

src/main/java/org/twittertip/CoSignedIdentifierAttestation.java
```diff
@@ -0,0 +1,81 @@
...
+            res.add(new DERBitString(signature));
+            return new DERSequence(res).getEncoded();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public boolean checkValidity() {
+        return getUnsignedAttestation().checkValidity();
```
jot2re, line 74: This method is never tested.

jot2re (changes requested):
Looks good, and follows the same pattern as `SignedIdentifierAttestation`. Only thing is that error handling and logging, could be done a bit more thorough like in
`SignedIdentifierAttestation`

PR #187, opened 2021-10-08
Updated attestation and cheque notValid computation 

jot2re:
It is not based on Timestamp, thus allowing for slack.
I also updated the default slack to 10 seconds instead of 1. 
I also updated tests that work on time validity to use allowed slack from Timestamp class to make them more robust to updates.


jot2re:
JS code looks good as well.

Issue #201, opened 2021-10-26
MetaMask signing of NFTAttestation looks weird


jot2re:
This is the back-end sister issue from the [alchemyNFT](https://github.com/AlphaWallet/AlchemyNFT/issues/143) repo.

jot2re:
Appropriate changes still need be made to the JS code when it becomes needed @oleggrib.
However, the Java code is ready. See https://github.com/TokenScript/attestation/pull/210 for detail.

jot2re:
This has laster been fixed in Authenticator. So I am closing this issue.

PR #202, opened 2021-10-29
MetaMask signing of NFTAttestation looks weird

jot2re:
This fixes issue #201


oleggrib (commented):
@jot2re code looks good. Js code updated accordingly.

Maybe worth to add some template for message hash? Like "Please sign message 'a8c755bb...' to prove that you own NFTs." or something similar. 

As option - we can send message template as input string to make it flexible.

jot2re:
I have added support for prefix/postfix message for the hash that gets signed. 
I also adjusted to have both a version 1 and 2 type of signature as per the discussion [here](https://github.com/AlphaWallet/AlchemyNFT/issues/143)

However, this PR is not sufficient by its own to make a new release. For this issue #203 should also be implemented. 
Once this is merged, I will continue working on issues #203.

jot2re:
@oleggrib I see that for some reason a conflict has occurred in the type script files. Can you check up on that and merge afterwards if you agree with everything?

jot2re:
Looks good

Issue #207, opened 2021-11-16
Create LisconNFT minter based on TokenAttestation object 


oleggrib:
Contract-related tasks:
0a. Our hosting server fire LisconNFT smartContract listener.
0b. We create LisconNFT smartContract image and save to the hosting ( @AW-STJ , can you ask someone to create image and contractMeta content? )
0c. We create LisconNFT smartContract metadata JSON and save to the hosting (@foxgem , can you provide URL what is OK for you?) 

How it will work:
1. User visit LisconTickets website 
2. LisconTickets Negotiator check if user have tickets, 
2a. no tickets. Message kind of: "Please check your mailbox and click magic link if you had a Liscon ticket."
3. Display tickets slider. 
4. User can click a ticket and code will create emailAttestation based on magicLink
5. Negotiator request Authenticator to create useToken object (ticket+emailAttestation+secretOwnershipProof) 
6. LisconTickets requests LisconNFT smartContract to validate useToken. 
7. LisconTickets sends useToken to the LisconNFT smartContract mint(useToken) function
8. LisconNFT sends useToken to ValidateAttestation smart contract to validate attestation. It returns [subjectAddress, ticketID, valid] ( @JamesSmartCell , are you working on that validator? can you do it? )
9. if object valid then LisconNFT mints token with id=ticketID and set subjectAddress as token owner  
10. Hosting Listener receive LisconNFT Transfer() event and create next files:
10a. generate TokenImage from template ( @AW-STJ , we need some template for token image. We can use tokenID to generate unique images, @foxgem , cant we use enpoint like you created for Derived contracts to save image/JSON ? )
10b. generate TokenMetadata JSON ( @AW-STJ , we need tokenMetaData template  )

ERC721 LisconNFT SmartContract code is ready. Website flow is ready too.

Most challenging part is create Validation Smart Contract.

@AW-STJ , what blockchain will we use for Liscon tokens? 
point me please to the Issue for that task in case if we already have one.

Any comments, updates, suggestions? @zhangzhongnan928 @colourful-land @JamesSmartCell ?

oleggrib:
Currently in Attestation lib for Attested Object we use EqualityProof. It use curvePoints substraction and other gas-expencive operations to validate proof.
Its ok for regualer server/browser, but too expencive for SmartContract

https://github.com/witnet/elliptic-curve-solidity
this is example lib for curvePoints manipulations in solidity . every operation costs lot of gas.
[...]

PR #208, opened 2021-11-19
Update contract and tests for Ticket AttestedObject

JamesSmartCell:
Update for the SmartContract handling for the Ticket AttestedObject.
Tests are now passing; should also include a negative test.

Linked Issue: #207


jot2re (changes requested):
It looks good! But as far as I can see only the ZKP is checked? 
I cannot seem to find the rest of the required checks done [here](https://github.com/TokenScript/attestation/blob/82cbe76d82d1df4a7c39a4873864a79264255d48/src/main/java/org/tokenscript/attestation/AttestedObject.java#L119) and [here](https://github.com/TokenScript/attestation/blob/82cbe76d82d1df4a7c39a4873864a79264255d48/src/main/java/org/tokenscript/attestation/AttestedObject.java#L153) being carried out.

JamesSmartCell:
> It looks good! But as far as I can see only the ZKP is checked? I cannot seem to find the rest of the required checks done [here](https://github.com/TokenScript/attestation/blob/82cbe76d82d1df4a7c39a4873864a79264255d48/src/main/java/org/tokenscript/attestation/AttestedObject.java#L119) and [here](https://github.com/TokenScript/attestation/blob/82cbe76d82d1df4a7c39a4873864a79264255d48/src/main/java/org/tokenscript/attestation/AttestedObject.java#L153) being carried out.

These are now all tested. I have added a minting test with lots of negative tests to ensure all functionality doesn't get broken.

There is also a deployment script.

The tests are in a TypeScript hardhat harness; this may need to be added to the Travis CI.

jot2re:
@JamesSmartCell I see the changes are thousands of lines so I think I need a pointer to the checking of the timestamp.
I would imagine there should be a function calling both the verification of the ZKP and timestamp and then returns the address of the issuer, attestor and subject where the issuer and attestor address then gets verified by the calling contract.

ethereum/attestation-test/hardhat.config.ts
```diff
@@ -81,7 +81,10 @@ export default {
       accounts: [`${PRIVATE_KEY}`]
     },
     mumbai: {
-      url: `https://matic-mumbai.chainstacklabs.com`, //ths RPC seems to work more consistently
+      url: `https://rpc-mumbai.maticvigil.com`, //ths RPC seems to work more consistently
```
hboon, line 84: There's also Infura:  `https://polygon-mumbai.infura.io/v3/#{key}`

oleggrib (approved):
@JamesSmartCell , it works good and we already use it. thanks

JamesSmartCell:
@jot2re since the code is in use now, safe to merge?

jot2re:
> @jot2re since the code is in use now, safe to merge?
@JamesSmartCell I would still like a pointer to the checks fo ZKP and timestamp, since it is really hard to find due to the amount of line changes.
Any pointers to the flow would be great. However, I can also verify this as part of the upcoming security review. But any kind of prose description of the specific smart contracts, their responsibility and interaction would greatly help me in this task and be much appreciated 😅

PR #210, opened 2021-11-26
Added eip712 signature version to signed NFT attestation 

jot2re:
This closes issue #201 
Furthermore this PR contains quite a bit of refactoring, renaming, added tests, and cleaning up on the files I have to touch in the process.

In general I refactored things so we can use any kind of ASN encoded object in an EIP712 signature with only a couple of lines of code, this makes it easy to update and generalise to other cryptographic objects.

The main things is that we now have two versions of SignedNFTAttestation, one based on the the approach we had before (constructing an ASN object containing the signature), the other obviously uses the EIP712 approach and shows the usual EIP712 fields and a short hex field for the underlying NFTAttestation to be signed.


jot2re:
Pushed the changes to main as they contain many refactoring. I ensured everything is backwards compatible, and validated that things work with attestation.id.

Issue #211, opened 2021-11-26
Add integer timestamps to attestation


jot2re:
Per request by @JamesSmartCell we need to add an extra field to our IdentifierAttesations containing an easily readable timestamp and expiration time. This reason is that currently it follows x509 and hence is a string. This costs a lot to parse in Solidity.

jot2re:
@JamesSmartCell @oleggrib 
The Java code is updated and merged with this feature. The JS and Solidity might need to be updated when needed. However, so far the Java code is fully backwards compatible with what is currently in place.

PR #212, opened 2021-11-28
Added optional timestamps in milliseconds

jot2re:
This closes issue #211 
Basically adds an optional encoding of the notBefore and notAfter timestamps of an Attestation as milliseconds since epoch instead of a Generalised Time string.
Unfortunately this means that the JS version must also be updated to handle Attestations including these extra timestamp encodings. 

This PR is per request of @JamesSmartCell for more efficient blockchain decoding.
Be aware that since we trust the issuer of the Attestation we do no expect inconsistencies between the x509 timestamp encoding and the ms since epoch encoding.


jot2re:
Merging to avoid staleness. Should be backwards compatible, as I tested the jar with attestation.id

Issue #221, opened 2021-12-21
Change Ticket_ID to String from integer 


AW-STJ:
Request from Devcon to change  ``ticket_id``  integer to a string. MIght be related to #115, but really to make it more flexible for future as well as to prevent a inelegant implementation of TicketID decoding for our current Liscon implementation.

jot2re:
@oleggrib @JamesSmartCell I will update this in Java, but I think I need your help to adjust the Solidity and JS versions accordingly. 
Furthermore, it is also important that we still have backwards compatibility and can understand the old tickets where the ID is an integer.

oleggrib:
@jot2re , we can set field devconId to variable( CHOISE ), so we can use Integer or String here. Its compatable with ASN1 so no problems here.
For JS we can read Integer/BigInt as string.
in case of solidity validation contract it need more work to update.
cc @colourful-land @JamesSmartCell

jot2re:
Yes, agree there is no issue when it comes to a complete ASN specification. 
Cool, that it should directly with JS.

jot2re:
I have made the java change in PR #225

jot2re:
@oleggrib please ensure that things also work in the JS 
@JamesSmartCell the smart contract code will also need to be updated to ensure understanding of tickets where the ticket ID is a string. It might be relevant to do this as a separate task though

oleggrib:
I just added commit for JS code. I think we can merge and @JamesSmartCell can update validation contract when we need to use it.

jot2re:
I have merged in the Java code, see PR #225, if we want to run with this change then the smart contract probably needs to be updated as well, but currently everything should be backward compatible @JamesSmartCell

JamesSmartCell:
Is this still pending?

JamesSmartCell:
> > Is this still pending?
> 
> As far as I can see, you handle it in #243, right?

Yes, that PR should close this issue.

PR #225, opened 2021-12-22
Change ticket ID to support string as well as integer

jot2re:
This closes issue #221


jot2re:
So far it has been implemented for java. 
It might need a small update for JS (or maybe it might actually out directly as mentioned by @oleggrib )

oleggrib:
JS code updated. Most of all Devcon will ask us to mint NFTs based on tickets, so we will need to update Validation Contract code to allow string or integer. @JamesSmartCell , can you update?

jot2re:
I am merging to avoid staleness. Everything should be backward compatible and I have tested the jar against attestation.id.

PR #227, opened 2022-01-10
Sonarqube review

jot2re:
I ran a static analysis tool on the attestation code and implemented the fixes it suggested to our code.


jot2re:
I am merging since this does not contain any functional changes but only increases robustness

PR #232, opened 2022-02-17
Add ZK annotation

JamesSmartCell:
Adding in information about the source of the Pedersen commitments for the ZK proofs.


weiwu-zhang:
@jot2re I think @JamesSmartCell's language is a bit off here, can you correct it?

He wrote:

````

    // Verify Proof Of Knowledge
    // forming a zk meta-proof that the two Pedersen commitments represent the same data
    // this verifies the integrity of the data; the actual data can be optionally revealed by the attestation holder
    // See:
    // Commitment1: https://github.com/TokenScript/attestation/blob/main/src/main/java/org/tokenscript/attestation/IdentifierAttestation.java
    // Commitment2: https://github.com/TokenScript/attestation/blob/main/src/main/java/org/devcon/ticket/Ticket.java
````

jot2re:
I enhanced the comments both in Solidity and Java

weiwu-zhang (approved):

PR #235, opened 2022-02-21
Change of the usage value for Liscon

jot2re:
Support for both legacy and new usage values have been done. 
A static construction method on `Eip712AttestationRequest` has been added to allow for decoded regardless of whether the old or new usage values have been used


oleggrib (approved):
It looks good, lets merge it.

PR #239, opened 2022-03-29
Javascript NPM packaging

micwallace:
Hey @jot2re & @oleggrib,

This branch adds stuff for publishing the JS as a typescript library on NPM (https://www.npmjs.com/package/@tokenscript/attestation)

It also:
- Moves postMesssage integration logic out of Authenticator to a separate IntegrationExample class for usage in the old demos. 
- Adds old schema versions, that are currently used by TN demos to a folder tn-compat. When the getUseTicket function is called there is an optional argument to use the old versions. These can be removed once we have updated TN ticket issuing to use the new schema. 
- Ensure ASN parsing errors are included in thrown exceptions in getUseTicket.

This doesn't necessarily need to be merged into main, we could keep a separate branch for JS library publishing and I'll keep it up to date with main. Let me know what you think. 

Cheers,
Michael


jot2re:
Hi Michael, thanks a lot! 
I think it would make sense to have it all merged in main to keep it all in place, and simply remove the code for handling tickets when we can. 
But @oleggrib is the one who is most aware of what is best to do regarding TS and JS code :)

jot2re:
@oleggrib can you check and merge this into main if you think it is fine?

micwallace:
Hey @oleggrib the library is now integrated with TN & attestation.id, and I have confirmed all existing tests are passing. 

Let me know if you want me to resolve the merge conflict from main.

jot2re (approved):
I have a hard time judging the TS/JS stuff, but ran a test and things still seem to work on my end.

src/main/javascript/crypto/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tokenscript/attestation",
-  "version": "1.0.7-alpha",
+  "version": "1.0.8-alpha",
```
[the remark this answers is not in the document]
micwallace: For major/minor parts of the semantic version I agree.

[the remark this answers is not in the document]
micwallace: This has been updated.

micwallace:
Hey @oleggrib & @jot2re 

As discussed [here](https://github.com/AlphaWallet/attestation.id/pull/127#issuecomment-1098066849) I've had to make the package into a commonjs version rather than es6 or hybrid to make it work with both TN and attestation.id. 

In the new version I have also added a proof validation function to Authenticator.js to validate the proof in TN. Can you have a look and see if it's okay?

jot2re:
Thanks a lot @micwallace ! I think we need @oleggrib to judge it

micwallace:
Hey @oleggrib & @jot2re 

Issues with the attestation packaging have been ironed out and it's working well as a commonjs module in attestation.id and TN for all platforms in use (webpack, UMD bundle, react, next.js). Please review the changes at your earliest convenience.

src/main/javascript/crypto/src/Authenticator.ts
```diff
@@ -52,106 +45,31 @@ const ALPHA_CONFIG = {
...
 
         // let ticket: Ticket = Ticket.fromBase64(base64ticket, KeyPair.fromPublicHex(base64senderPublicKey));
         try {
-            ticket = Ticket.fromBase64(base64ticket,{"6": KeyPair.publicFromBase64(base64senderPublicKey)});
+            if (useOldTicketSchema){
+                ticket = TicketTNCompat.fromBase64(base64ticket,KeyPair.publicFromBase64(base64senderPublicKey));
+            } else {
+                ticket = Ticket.fromBase64(base64ticket,{"6": KeyPair.publicFromBase64(base64senderPublicKey)});
+            }
+
```
oleggrib, line 72: Better remove hardcoded conferenceID "6" and receive keysArray as argument, like here
https://github.com/TokenScript/attestation/blob/ticket_validator/src/main/javascript/crypto/src/Authenticator.ts#L682-L710

src/main/javascript/crypto/src/Authenticator.ts
```diff
@@ -163,7 +81,7 @@ export class Authenticator {
             logger(DEBUGLEVEL.MEDIUM,'ticked valid (signature OK)');
         } catch (e) {
             logger(DEBUGLEVEL.MEDIUM,'getUseTicket: ticket validation failed',e);
-            throw new Error("getUseTicket: ticked validation failed");
+            throw new Error("getUseTicket: ticked validation failed: " + e.message);
```
oleggrib, line 84: we need some way to set DEBUGLEVEL, because it can't be set in config.ts, because its will be ready NPM package.
lets add some variable to the .env , so code can read and use it

src/main/javascript/crypto/src/libs/AttestedObject.ts
```diff
@@ -146,13 +146,12 @@ export class AttestedObject implements ASNEncodable, Verifiable {
         pok.fromBytes( new Uint8Array(attested.proof) ) ;
         me.pok = pok;
 
-        let attCom: Uint8Array = me.att.getUnsignedAttestation().getCommitment();
-        let objCom: Uint8Array = me.attestableObject.getCommitment();
-        let crypto = new AttestationCrypto();
+        me.userKeyPair = me.att.getUnsignedAttestation().getSubjectPublicKeyInfo();
```
oleggrib, line 150: @micwallace , why do you remove this check?

src/main/javascript/crypto/src/tn-compat/SignedDevconTicket.ts
```diff
@@ -0,0 +1,29 @@
+import { AsnProp, AsnPropTypes } from "@peculiar/asn1-schema";
+import { ValidityValue, AlgorithmIdentifierASN} from "../asn1/shemas/AuthenticationFramework";
+
+export class DevconTicket {
+    @AsnProp({ type: AsnPropTypes.Utf8String }) public devconId: string;
+    @AsnProp({ type: AsnPropTypes.Integer }) public ticketId: number;
+    @AsnProp({ type: AsnPropTypes.Integer }) public ticketClass: number;
+}
```
oleggrib, line 8: Can you check main branch for code?
https://github.com/TokenScript/attestation/blob/main/src/main/javascript/crypto/src/asn1/shemas/SignedDevconTicket.ts#L10-L17
its single object , compatible with old and new ticket ASN schema
and Ticket.ts updated to be compatable with ID as integer/string and old and new ticket formats

oleggrib (commented):
Can you check my comments and make some updates/comments so we can move this repo to the /authenticator and avoid duplication.
for testing - we can put /attestation and /authenticator next each to other and update path to read JAVA-builds to test JS version.
@micwallace , @jot2re , ok?

data-modules/output/asn/AuthenticationFramework.asn
```diff
@@ -19,9 +19,9 @@ Validity ::= CHOICE {
 		
 ValidityValue ::= SEQUENCE {
     notBefore       GeneralizedTime,
-    notBeforeInt    INTEGER OPTIONAL, -- NotBefore time since epoch in milliseconds
+    notBeforeInt    INTEGER OPTIONAL, -- NotBefore time since epoch in seconds
     notAfter        GeneralizedTime,
-    notAfterInt     INTEGER OPTIONAL -- NotAfter time since epoch in milliseconds
+    notAfterInt     INTEGER OPTIONAL -- NotAfter time since epoch in seconds
```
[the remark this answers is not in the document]
jot2re: Yes, they are definitely needed for something running on the smart contract. However, optional for the any web2 usage. Because if we make them all required then the schema is not backward compatible with old attestations (furthermore GeneralizedTime becomes totally superfluous and should just be removed).
However, as we talked about loosing backward compatibility is not really an issue since we have only issued 1 hour valid attestations for awhile, and I think it is fair to deprecated those few "unlimited time" old attestations we issued. 
So ideally I would just want to remove GeneralizedTime and deprecate the initial attestation, but I think that should be cleared with Weiwu. This would also require many code changes.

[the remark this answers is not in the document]
jot2re: The logic in all the code bases (including smart contract), is that notAfterInt is optional but required to work on smart contracts, and when new attestations are issued notAfterInt and notBeforeInt _will always_ be set unless explicitly requested to not be included.

src/main/javascript/crypto/src/libs/Attestation.ts
```diff
@@ -64,6 +65,13 @@ export class Attestation {
         if (decodedAttestationObj.validity){
             me.notValidBefore = decodedAttestationObj.validity.notBefore.generalizedTime.getTime();
             me.notValidAfter = decodedAttestationObj.validity.notAfter.generalizedTime.getTime();
+            // TODO validate time when it will be updated in Java code
```
jot2re, line 68: It should be updated in the java code, so I think the code fragment can be enabled.

src/main/javascript/crypto/src/IntegrationExample.ts
```diff
@@ -0,0 +1,203 @@
...
+        //     identifierAttestation: identifierAttestation,
+        //     proof: proof
+        // })
+        // // Serialise it (for use as a transaction parameter) and return it
+        // return useDevconTicket.serialize();
+    }
+
+    attachPostMessageListener(listener: Function){
+        if (window.addEventListener) {
+            window.addEventListener("message", (e) => {
```
jot2re, line 192: To avoid potential cross origin attacks, it should be validated that the origin of the message is as expected, ie. something like the following should be added
``` 
if (event.origin !== "http://example.org") // Compliant
    return;
```

data-modules/output/asn/AuthenticationFramework.asn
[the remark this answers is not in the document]
oleggrib: yes, correct, but do we use option to force old legasy format somewhere? As I know - we dont use it and always add int values.
even more - libs updated to always read Int and GeneralizedTime both(if exists) and compare them, so in real usage we always use new format only @jot2re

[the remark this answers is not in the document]
jot2re: Yes, true. I do think the best would simply be to get rid of the GeneralizedTime field. But I don't think we should do that before we get a good and integrated CI/CD setup.
I can change the backend code to _always_ enforce that notBeforeInt and notAfterInt are included, but in that case we should make a separate issue, since this is not related to NPM packaging ;)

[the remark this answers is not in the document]
oleggrib: ok, lets leave it for separate issue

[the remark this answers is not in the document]
jot2re: Added a new issue #249

src/main/javascript/crypto/src/tn-compat/Ticket.ts
```diff
@@ -0,0 +1,174 @@
...
+        this.createWithCommitment(devconId, ticketId, ticketClassInt, new Uint8Array(commitment), uint8tohex(new Uint8Array(signature)) , keys );
+    }
+
+    public getCommitment(): Uint8Array {
+        return this.commitment;
+    }
+
+
+    public getUrlEncoding() {
+        // TODO implement
```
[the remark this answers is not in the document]
oleggrib: usually throw an error is good option, but in this case I prefer to use Ticket reader from main branch, so no need to update duplicated code. we can remove tn-compat/Ticket.ts because schema is not compatable with legacy ticket. main branch already can work with both ticket types.

src/main/javascript/crypto/src/IntegrationExample.ts
oleggrib: this part of code just add event listener , so we have to check event.origin in the callback function. this part of code doesnt have access to the event object

src/main/javascript/crypto/src/tn-compat/SignedDevconTicket.ts
oleggrib: we can merge this branch with main and stay Tickets.ts from main branch. it can read both Ticket formats

src/main/javascript/crypto/src/Authenticator.ts
oleggrib: this code can be updated after merge with main, because its related to the Ticket

src/main/javascript/crypto/src/Authenticator.ts
oleggrib: added commit, where I moved debug variable to the process.env.DISPLAY_DEBUG_LEVEL or window.DISPLAY_DEBUG_LEVEL so we can change debug level even for UMD builds

PR #243, opened 2022-04-06
Test Ticket Validity time window

### LINKED ISSUE #241: Update attestation validation smartContract
Hi, @JamesSmartCell 

Can you look at https://github.com/TokenScript/attestation/pull/212
in this PR Tore have added optional timestamp as unixtime in seconds, so attestation schema is changed.

Can you update attestation validation smartContract to validate timestamp and make it compatible with last attestation schema?

https://github.com/TokenScript/attestation/pull/237 (JS updates)

JamesSmartCell:
Added handling of Blockchain friendly timestamps now included in the Attestation format.

Notes:

- For testing it was required to make the Attestation timestamp validity window slightly pre-dated because the ethereum timestamp used is the block timestamp which is only updated every block. If you create the attestation after a block has been written and check validity immediately the test will almost always fail (there would be an edge condition where you create the attestation right before the block is written, and the block is written while you are sending the query to the node).
- The time check doesn't fail when given a non-blockchain friendly timestamp (without the integer unix time) but it will always pass in this case, the timestamp is not checked.

Should timestamp validity fail if the blockchain friendly timestamp is not used?

Closes #241


jot2re:
> Should timestamp validity fail if the blockchain friendly timestamp is not used?

I am actually not sure what the best flow would be. Do we have something issued from already that we expect to actually be used? If yes, then it the parsing should happen on the non-blockchain friendly timestamp (although I am not sure you have convenient library handles to do that).
If no, then the parsing should fail.

@JamesSmartCell what do you think? Can we assume that no previously issued Tickets will actually used on the blockchain?

jot2re:
I think something better, is simply to do the following flow: If blockchain friendly timestamp is not there, ensure that conference ID is "6". If so, accept, otherwise fail.

JamesSmartCell:
Hi @jot2re what is the conferenceID of 6? This is just for the comments in the source code so it makes sense to our future selves.

oleggrib:
> I think something better, is simply to do the following flow: If blockchain friendly timestamp is not there, ensure that conference ID is "6". If so, accept, otherwise fail.

but attestation its separate object from ticket. 
from now all attestations will have blockchain friendly timestamp

all previous attestation expired already

JamesSmartCell:
> > I think something better, is simply to do the following flow: If blockchain friendly timestamp is not there, ensure that conference ID is "6". If so, accept, otherwise fail.
> 
> but attestation its separate object from ticket. from now all attestations will have blockchain friendly timestamp
> 
> all previous attestation expired already

Does that mean it's safe to remove that check from the contract?

JamesSmartCell:
@jot2re @oleggrib contract is updated to handle BlockchainFriendly time in epochseconds, matching with the new update to Attestation.

JamesSmartCell:
Why do all the tests fail? Locally they work correctly.

jot2re:
> Why do all the tests fail? Locally they work correctly.

The issue is in commit 3a1b1bd478acb2e83d12815507c691050dc84330 
I think it might have resulted from a merge conflict or something like that. Basically the integer timestamp got multiplied with 1000. 
I have fixed it now

jot2re:
> > > I think something better, is simply to do the following flow: If blockchain friendly timestamp is not there, ensure that conference ID is "6". If so, accept, otherwise fail.
> > 
> > 
> > but attestation its separate object from ticket. from now all attestations will have blockchain friendly timestamp
> > all previous attestation expired already
> 
> Does that mean it's safe to remove that check from the contract?

@JamesSmartCell yes, I just with @oleggrib about this. While we at some point issued some attestations with unlimited validity, we both agree that we can simply reject *any* attestation that is not blockchain friendly because the user can always get a new one issued.
So if you could update the contract to always fail if the attestation is not blockchain friendly, that would be great

ethereum/attestation-test/contracts/VerifyTicket.sol
```diff
@@ -200,18 +193,52 @@ contract VerifyTicket {
...
+        (length, timeBlock, decodeIndex, tag) = decodeElement(attestation, decodeIndex + length); //INTEGER_TAG if blockchain friendly time is used
+        if (tag == INTEGER_TAG)
+        {
+            uint256 startTime = bytesToUint(timeBlock);
+            (length, decodeIndex, ) = decodeLength(attestation, decodeIndex); //18 0F
+            (, timeBlock, decodeIndex,) = decodeElement(attestation, decodeIndex + length);
+            uint256 endTime = bytesToUint(timeBlock);
+            valid = block.timestamp > startTime && block.timestamp < endTime;
+        }
+        else
```
jot2re, line 232: After discussion with @oleggrib we agreed that the best is simply to fail if the attestation does not contain a blockchain friendly timestamp.

ethereum/attestation-test/contracts/VerifyTicket.sol
```diff
@@ -79,46 +84,34 @@ contract VerifyTicket {
...
         //ensure the attestor and issuer keys are correct
         address recoveredAttestor;
-        address recoveredIsuer;
+        address recoveredIssuer;
 
-        (recoveredAttestor, recoveredIsuer, subject, ticketId) = verifyTicketAttestation(attestation);
+        (recoveredAttestor, recoveredIssuer, subject, ticketId, conferenceId, timeStampValid) = verifyTicketAttestation(attestation);
 
-        if (recoveredAttestor != attestor || recoveredIsuer != ticketIssuer)
+        if (recoveredAttestor != attestor || recoveredIssuer != ticketIssuer || !timeStampValid)
```
jot2re, line 107: Maybe this gets verified by the calling contract? But it is _crucial_ to validate that the `subject` returned here is the same as the key used to sign the transaction.

src/test/java/org/devcon/ticket/UseTicketTest.java
```diff
@@ -150,19 +150,59 @@ public void testDecoding() throws InvalidObjectException {
...
+    att = HelperTest.makeUnsignedStandardAtt(subjectKeys.getPublic(), ATTESTATION_SECRET, MAIL, 15); //valid for 15 seconds
+    att.blockchainFriendly = false;
+    signed = new SignedIdentifierAttestation(att, attestorKeys);
+    ticket = new Ticket(MAIL, "7", TICKET_ID, TICKET_CLASS, ticketIssuerKeys, TICKET_SECRET); //using conferenceId 6 should pass (legacy test)
+    useTicket = new AttestedObject<>(ticket, signed, ATTESTATION_SECRET, TICKET_SECRET, UN, crypto);
+
+    //test should fail:
+    tar = contract.callVerifyTicketAttestation(useTicket.getDerEncoding());
+    assertFalse(tar.timeStampValid);
+
```
jot2re, line 204: We should also have a tests that fail in case of
1. Wrong key used to sign the transaction.
2. Wrong Ticket issuer key is used
3. Wrong attestor key used

jot2re (changes requested):
Looks good, but I think there a couple of things that must still be handled as mentioned in the in-line comments.

Furthermore, one other thing is that  is also a VerifyTicket.sol in demo which is different from the one in ethereum/attestation-test.

Are both necessary?

src/test/java/org/devcon/ticket/UseTicketTest.java
```diff
@@ -150,19 +150,59 @@ public void testDecoding() throws InvalidObjectException {
   @Test
   public void testSmartContractDecode() throws Exception {
     //try building all components
-    IdentifierAttestation att = HelperTest.makeUnsignedStandardAtt(subjectKeys.getPublic(), ATTESTATION_SECRET, MAIL);
+    IdentifierAttestation att = HelperTest.makeUnsignedStandardAtt(subjectKeys.getPublic(), ATTESTATION_SECRET, MAIL, 15); //valid for 15 seconds
```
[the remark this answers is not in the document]
jot2re: TBH it is just a bit annoying to have to wait 30 sec for a test when you run the unit tests often... as I have done the last couple of days 😅

ethereum/attestation-test/contracts/VerifyTicket.sol
JamesSmartCell: > Maybe this gets verified by the calling contract? But it is _crucial_ to validate that the `subject` returned here is the same as the key used to sign the transaction.

Does the subject sign? In this attestation there's no signature from the subject (at least as far as I can see), only their public key. We have issuer and attestor signatures, which are checked.

src/test/java/org/devcon/ticket/UseTicketTest.java
JamesSmartCell: We can test (2) and (3).

I agree we need a test for (1) too - I can add this into the HardHat .ts suite as it involves setting up a function which takes the attestation as payload, sending a transaction which calls that function and checks the result.

JamesSmartCell:
Ok I'll make the small update to the contract and add the additional tests.

ethereum/attestation-test/contracts/VerifyTicket.sol
jot2re: Okay, I see what you mean. But then we put a lot of responsibility on the external library to ensure that they use the address of `subject` to potentially transfer tokens to instead of the address that signed the request.

But I understand your point. Maybe we can and a comment in nice bold letters? :)

src/test/java/org/devcon/ticket/UseTicketTest.java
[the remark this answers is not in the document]
JamesSmartCell: > TBH it is just a bit annoying to have to wait 30 sec for a test when you run the unit tests often... as I have done the last couple of days 😅

Ok then, now we know it works ok we can 'cheat' and build a slightly out of date attestation.

ethereum/attestation-test/contracts/VerifyTicket.sol
JamesSmartCell: > Okay, I see what you mean. But then we put a lot of responsibility on the external library to ensure that they use the address of `subject` to potentially transfer tokens to instead of the address that signed the request.
> 
> But I understand your point. Maybe we can and a comment in nice bold letters? :)

Ok I'll add that comment. 

Sometimes we might be calling the transaction from a server as a curtesy so the user doesn't need to spend gas (if they don't already have eth), in which case the transaction sender will be different from the destination of eg a minted token.

src/test/java/org/devcon/ticket/UseTicketTest.java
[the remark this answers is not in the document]
JamesSmartCell: > TBH it is just a bit annoying to have to wait 30 sec for a test when you run the unit tests often... as I have done the last couple of days 😅

I had to make a modification to the ALLOWED_ROUNDING to accommodate a 19 seconds in the past expiry of an attestation.

JamesSmartCell:
Hi @jot2re, have done the refactoring:

- Use standard Interface for Verification contract.
- Always fail non-blockchain friendly timestamps.
- Use attestation which expired 19 seconds in the past for invalid timestamp test (deal with Ethereum block time update).

Added these tests:

- Check ticket subject is correct (in the HardHat test, since it involves sending transaction).
- Negative tests with incorrect issuer and attestor.

JamesSmartCell:
@jot2re Sorry I had to slightly change the value of ```ALLOWED_ROUNDING``` from 10000 to 20000 to be able to set the expiry 19 seconds in the past. This meant I needed to add a 10000 value into the TimeStampTest in order for it to pass.

ethereum/attestation-test/contracts/VerifyTicket.sol
jot2re: > > Okay, I see what you mean. But then we put a lot of responsibility on the external library to ensure that they use the address of `subject` to potentially transfer tokens to instead of the address that signed the request.
> > But I understand your point. Maybe we can and a comment in nice bold letters? :)
> 
> Ok I'll add that comment.
> 
> Sometimes we might be calling the transaction from a server as a curtesy so the user doesn't need to spend gas (if they don't already have eth), in which case the transaction sender will be different from the destination of eg a minted token.

Yes, after your previous reply I realised that not validating the origin of the transaction matching the subject, could be considered a feature. So I think you are right, that it is best _not_ to check this in the smart contract, but instead document it.

jot2re (changes requested):
I looked a bit further into the tests and what got returned and hence I got a couple of further inline comments of minor changes that I think would be good. Sorry I didn't notice these before.

I also expanded a bit on the tests and found that the failures did not always occur as expected. I do find this very surprising because when I read the Solidity code I would expect the tests to pass as they are now.
Please pull and have a look at the out-commented lines `UseTicketTest`.

ethereum/attestation-test/contracts/VerifyTicket.sol
```diff
@@ -91,18 +91,16 @@ contract VerifyTicket {
         uint length;
     }
 
+    /**
+    * Perform TicketAttestation verification
+    * NOTE: This function DOES NOT VALIDATE the subject; you must perform validation of the subject from the calling function.
+    **/
     function verifyTicketAttestation(bytes memory attestation, address attestor, address ticketIssuer) external view returns(address subject, bytes memory ticketId, bytes memory conferenceId, bool timeStampValid)
```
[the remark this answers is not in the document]
JamesSmartCell: > I think it makes more sense to simply return a single value indicating whether validation is ok or not, rather than only a boolean indicating whether the timestamp is valid or not

We still need to return the subject, ticket and conference id for any calling contract, do you mean that the boolean should be a global 'pass/fail' of the whole attestation?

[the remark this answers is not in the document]
jot2re: Yes, sorry, that is exactly what I mean. The Boolean should indicate whether **anything** went wrong

ethereum/attestation-test/contracts/VerifyTicket.sol
```diff
@@ -125,7 +128,7 @@ contract VerifyTicket {
 
         (ticketIssuer, ticketId, conferenceId, commitment2, decodeIndex) = recoverTicketSignatureAddress(attestation, decodeIndex);
 
-        (attestor, subject, commitment1, decodeIndex, timeStampValid) = recoverSignedIdentifierAddress(attestation, decodeIndex, conferenceId);
+        (attestor, subject, commitment1, decodeIndex, timeStampValid) = recoverSignedIdentifierAddress(attestation, decodeIndex);
 
         //now pull ZK (Zero-Knowledge) POK (Proof Of Knowledge) data
         (pok, decodeIndex) = recoverPOK(attestation, decodeIndex);
```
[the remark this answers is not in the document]
JamesSmartCell: Updated

ethereum/attestation-test/contracts/VerifyTicket.sol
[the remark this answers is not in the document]
JamesSmartCell: Updated

jot2re (approved):
Looks great to me!

jot2re (changes requested):
I was a bit too quick at approving. I realised the update breaks the hardhat test.

ethereum/attestation-test/contracts/VerifyTicket.sol
oleggrib: @jot2re , in this case better make validityInteger as required value, not optional in the Java lib.

jot2re:
@JamesSmartCell I tried to test with the latest changes, but I unfortunately still have 4 failing test in relation to minting NFTs smart contract functions. Do you still have any failures when you run it locally @JamesSmartCell ?

ethereum/attestation-test/contracts/VerifyTicket.sol
jot2re: Good point! Currently we always include it. The optionality is only to reflect that the old attestations technically are correct (and should work in the browser), such that the spec is backward compatible. But I am changing things now, so that it is not possible to set it to false when constructing  new attestations, but the old ones can still be decoded.

JamesSmartCell:
The Liscon specific tests fail because those weren't finished when I branched. I'll check with @oleggrib

JamesSmartCell:
@jot2re can this be merged yet? We could remove the Liscon tests which must fail because they are pre-generated with a 1 hour time window.

jot2re:
Yes, since with me. I made an issue #254 , which discusses redoing these integration tests and adding other integration tests as part of our goal of improving CI/CD.

jot2re (approved):

Issue #246, opened 2022-04-15
Constant time crypto operations


jot2re:
Currently the crypto implementation in both backend and frontend is not constant time, since it didn't really seem like a feasible attack model. However, for completeness and because ECDSA is so sensitive to timing attacks it is better to make it constant time.

One could imagine an attacks of someone trying to continuously querying attestation.id for certificates on emails and time the response time to try to learn information about the secret signing key.

weiwu-zhang:
closed per #248

PR #248, opened 2022-04-20
246 constanttime

jot2re:
this closes issue #246


weiwu-zhang:
I didn't read the PR, but think there is no reason not to approve it.

PR #257, opened 2022-05-10
Fix ASN parse buffer conversion.

micwallace:
- Remove redundant and problematic conversion from uInt8Array to buffer.
  UInt8Array is already a BufferSource that the ASNParse.parse method accepts.
  Removing this conversion prevents the error in token negotiator tests:
  Wrong parameter: inputBuffer must be "ArrayBuffer"

- Jest updated to handle native support of UInt8Array.


jot2re:
@oleggrib can you check this PR?

oleggrib:
@micwallace , ASN parser define : 
```export declare type BufferSource = ArrayBuffer | ArrayBufferView;```
so UInt8Array is not an ArrayBuffer

we have to change UInt8Array to ArrayBuffer for AsnParser. 

Could you more detailed describe conversion problem? maybe we can convert it in other way?

micwallace:
Hey @oleggrib UInt8Array is an ArrayBuffer view. The buffer can be accessed simply by using uInt8Array.buffer.

This is done in the ASN parser code here: https://github.com/PeculiarVentures/asn1-schema/blob/78eea4acb621fea1e67b7a1d433ee4afac8f803d/packages/schema/src/parser.ts#L27

The problem with the previous code is that it's using the node.js buffer library. This is usually fine because it's polyfilled, but for some reason was causing an error with jest tests for token-negotiator. We should try to avoid node.js core libraries where possible as they can cause issues and often require the usage of outdated or unmaintained polyfills.

jot2re:
I just ran the tests to validate things, unfortunately I get a failure when running `npm run testjest`.
Specifically an error occurs in a test which uses the method `KeyPair.ts:publicFromPEM(pem: string): KeyPair`.

@micwallace @oleggrib could one of you have a look at this and see if you can reproduce?

micwallace:
Hey Tore,

I had the same issue when I ran the tests today but this fixed it:

1. Delete node_modules folder
2. Delete package-lock.json
3. Run npm install

Can you try this and see if tests pass?

On 16 May 2022 6:23:24 pm AEST, Tore Frederiksen ***@***.***> wrote:
>I just ran the tests to validate things, unfortunately I get a failure when running `npm run testjest`.
>Specifically an error occurs in a test which uses the method `KeyPair.ts:publicFromPEM(pem: string): KeyPair`.
>
***@***.*** @oleggrib could one of you have a look at this and see if you can reproduce?
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/TokenScript/attestation/pull/257#issuecomment-1127376032
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

oleggrib:
> I just ran the tests to validate things, unfortunately I get a failure when running `npm run testjest`. Specifically an error occurs in a test which uses the method `KeyPair.ts:publicFromPEM(pem: string): KeyPair`.

or just run ```npm i @peculiar/asn1-schema``` , it should solve the issue

PR #265, opened 2022-05-19
updated logging, error handling, added method to read key from PEM st…

oleggrib:
…ring


micwallace (approved):
Awesome PR Oleg! I'll test against att.id and TN tomorrow and publish a new build.

PR #273, opened 2022-06-23
fix: 🐛 asn_decoding fix(patch), attestation timestamp check

oleggrib:
Added patch to fix ASN1_schema bug for Integers, added Attestation
validate timestamp

### DIFF
--- a/src/main/javascript/crypto/package.json
+++ b/src/main/javascript/crypto/package.json
@@ -12,7 +12,8 @@
     "clean": "rm -R dist",
     "watch": "webpack --watch --progress",
     "testjest": "jest",
-    "prepublishOnly": "npm run clean && npm run build"
+    "prepublishOnly": "npm run clean && npm run build",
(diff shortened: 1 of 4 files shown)


micwallace:
@oleggrib I'm not sure parsing to string is the best method, as it could result in being an integer unexpectedly for smaller integers. It may be better to always return a number for consistency. Or return a big int if bigger than 8 bytes. The ASN specification doesn't specify a maximum value for integer fields. So number should be fine for anything up to 64bits and big int for the rest.

oleggrib:
@micwallace , by the way - I created an issue for ASN lib and Author fixed the bug, so we can avoid patch and use last version of lib. but it still have option to return string

https://github.com/PeculiarVentures/asn1-schema/blob/master/packages/schema/src/converters.ts#L30-L35

so we have to handle it too

micwallace (approved):

PR #274, opened 2022-06-28
feat: 🎸 js implment for issue 246

oleggrib:
issue 246: java PR#248, validate point against curve

✅ Closes: #246

@jot2re , can you check what is wrong with code : 
https://github.com/TokenScript/attestation/blob/246-js-constanttime/src/main/javascript/crypto/src/libs/AttestationCrypto.ts#L389-L393

sometimes it gets symetric points, but maybe because of missing normalization it breaks. Maybe you can help to solve it.

or help me to understand reason of ```point.multiplyDA(curve.n).isInfinity()```

### DIFF
--- a/src/main/javascript/crypto/src/asn1/shemas/miccy.ts
+++ b/src/main/javascript/crypto/src/asn1/shemas/miccy.ts
@@ -1,0 +1,16 @@
+import {AsnArray, AsnProp, AsnPropTypes, AsnType, AsnTypeTypes} from "@peculiar/asn1-schema";
+
+
+export class EpochTimeValidity {
+    @AsnProp({ type: AsnPropTypes.Integer }) public notBefore: number;
+    @AsnProp({ type: AsnPropTypes.Integer }) public notAfter: number;
+}
+
+export class SimpleAsnIntSchema {
+    @AsnProp({ type: EpochTimeValidity }) public validity: EpochTimeValidity;
+}
+
+export class s2 {
+    @AsnProp({ type: AsnPropTypes.Integer }) public notBefore: number;
+}
+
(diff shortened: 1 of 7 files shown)


jot2re:
@oleggrib that is a very good question! Normalisation should not have any influence here (normalisation only has something to do with internal representation of a point on a curve). Basically when the check fails it mathematically means that the point does not have the correct order. In non-math terms it means that the point is not safe to use. 
This could indicate a bug somewhere either in computation of points or decoding of points.
Does it fail when it uses data made with attestation.jar, or does the failure only happen when it decoded something constructed by authenticator.jar?

jot2re:
I had a quick check on the Authenticator code and I could not find anything wrong with your code. The risk is that one of the libraries used might contain a bug.

jot2re:
Looking a bit closer it might be an issue with the code in `Point.ts`. It looks like you normally represent the point at infinity as either `x=null` or `y=null`. But in the implementation of `multiplyDA` it looks like you use `x=0` and `y=0` to represent the point at infinity through the command `newZero`.

oleggrib:
> Does it fail when it uses data made with attestation.jar, or does the failure only happen when it decoded something constructed by authenticator.jar?

it fails for example in next tests:
MagicLink reader › Decode Magic Link from Java Build
MagicLink reader › Encode/Decode Magic Link from JS

so in both cases it failed.

how it fails:
https://github.com/TokenScript/attestation/blob/246-js-constanttime/src/main/javascript/crypto/src/libs/AttestationCrypto.ts#L390
->
https://github.com/TokenScript/attestation/blob/246-js-constanttime/src/main/javascript/crypto/src/libs/Point.ts#L114
->
https://github.com/TokenScript/attestation/blob/8468dac0d749ae5e55e61eed5d89db84e2b7ce6e/src/main/javascript/crypto/src/libs/Point.ts#L102 (by logic it can be a this.newZero(), but our point have positive Y2, so Y1 != -Y2)
-> then code try to invert point where X1 - X2 == 0 , but Y1 == -(Y2 - module) and fails

maybe it can help you to understand how to solve the issue. @jot2re

oleggrib:
thank you, @jot2re , it looks and works great, I thought it something simple, so when error happened it was already correct result, but wrong check happened.

So we can merge this PR , close issue and remove branches. Can you do it, @jot2re if it looks good for you?

jot2re (approved):
I checked and everything looks good. There were a couple of checks missing which I added. Everything runs locally on my machine.

PR #279, opened 2022-07-04
feat: ✨ safeconnect

jot2re:
Backend code for Safe connect


micwallace:
@jot2re I think the safe connect code is at the point where we can merge. There may be a few minor changes in the future but the core if it is in place and it doesn't effect other features in the library. What do you think?

jot2re:
@micwallace yes, I agree. Better to merge before too many other changes happens on `main`.
Let me just check through the latest changes

jot2re:
I am merging by agreement with @micwallace to avoid staleness. 
At some point though, it would be great to get @oleggrib to have a look at the new safe connect packages in both JS and Java.
I have made a Jira issue for (this)[https://smarttokenlabs.atlassian.net/jira/software/projects/TKN/boards/5?selectedIssue=TKN-381]

Issue #276, opened 2022-06-29
Eip712Validator.ts is lacking some checks


jot2re:
In Eip712Validator.ts.validateRequest()
Eip712AttesationRequest.ts.checkValidity()
Eip712AttesationRequestWithUsage.ts.checkValidity()
token-authenticator does not validate chainID, verifying contract or salt, whereas this is validated in the equivalent classes in attestation.jar. It means that signed requests can be used on any contract or chain in the future. 

See section 2.2.3 in the Token-negotiator report.
See (Jira issue 291)[https://smarttokenlabs.atlassian.net/browse/PR-291].

PR #281, opened 2022-07-08
fix: 🐛 issue#276 fix. Added eip712 domain validations

oleggrib:
✅ Closes: #276

### DIFF
--- a/src/main/javascript/crypto/src/Authenticator.ts
+++ b/src/main/javascript/crypto/src/Authenticator.ts
@@ -197,6 +197,7 @@ export class Authenticator {
         let pok = crypto.computeAttestationProof(secret, nonce);
         let attRequest = AttestationRequest.fromData(crypto.getType(type), pok);
         let attestationRequest = new Eip712AttestationRequest(userKey);
+        // undefined means that we will use default value in attestationRequest.addData() second parameter
         await attestationRequest.addData(attestorDomain, undefined, receiverId, attRequest);
 
         Authenticator.checkAttestRequestVerifiability(attestationRequest);
--- a/src/main/javascript/crypto/src/Authenticator.ts
+++ b/src/main/javascript/crypto/src/Authenticator.ts
@@ -124,7 +124,6 @@ export class Authenticator {
         }
         logger(DEBUGLEVEL.HIGH,'attestation valid');
         
-
         try {
             let redeem: AttestedObject = new AttestedObject();
             redeem.create(ticket, att,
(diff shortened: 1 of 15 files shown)


src/main/javascript/crypto/src/Authenticator.ts
```diff
@@ -197,9 +196,13 @@ export class Authenticator {
 
         let pok = crypto.computeAttestationProof(secret, nonce);
         let attRequest = AttestationRequest.fromData(crypto.getType(type), pok);
-        let attest = new Eip712AttestationRequest(userKey);
-        await attest.addData(attestorDomain, 20*1000, receiverId, attRequest);
-        let attestJson = attest.getJsonEncoding();
+        let attestationRequest = new Eip712AttestationRequest(userKey);
+        await attestationRequest.addData(attestorDomain, undefined, receiverId, attRequest);
```
jot2re, line 200: Why `undefined` validity? How exactly is that encoded? Shouldn't it be default validity?

jot2re (changes requested):
Thanks for the update! I think there might still be a couple of checks that should be added. 
In particular I think that some updates are also needed to the validation in Eip712AttestationRequestWithUsage and Eip712AttesationUsage in the sense that they should perform the same domain validation as Eip712AttesationRequest.

src/main/javascript/crypto/src/Authenticator.ts
oleggrib: when function receive "undefined" then it use default value

src/main/javascript/crypto/src/libs/Eip712Validator.ts
```diff
@@ -42,56 +46,61 @@ export class Eip712Validator {
         return this.domain;
     }
 
-    validateRequest(jsonInput: string) {
-        try {
-            let authenticationData = JSON.parse(jsonInput);
+    // we dont use it in Java code
```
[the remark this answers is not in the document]
oleggrib: removed

jot2re:
The issue of missing domain validation turned out to also be present in attestation.jar. I handled this in [this PR](https://github.com/TokenScript/attestation/pull/285)

jot2re:
I also implemented the JS fix in [this commit ](https://github.com/TokenScript/attestation/pull/281/commits/8c5da8a2a5786ee36602c5b18ab359f97f5e9174). So I am happy with merging if you are @oleggrib

jot2re (approved):

Issue #275, opened 2022-06-29
Make the UN based on a signature instead of a MAC


jot2re:
Currently the construction of a UN is based on a MAC. This works fine in the online setting, but we are getting into situations where we need to be able to do validate offline. In particular in the blockchain.
For this reason it makes sense to change the structure of the unpredictable number to be based on a digital signature (ECDSA secp256k1) instead of a MAC.

jot2re:
In relation to this and the current UN implementation I think we should increase it's length from 64 bits to 128 bits since it now  not only work to give randomness but also authenticity of the randomness.

jot2re:
Closed with PR #284

Issue #259, opened 2022-05-13
Add option of including context information in construction of UN


jot2re:
Currently a UN can be used multiple times within its validity. 
This is a feature that allows it to be used repeatedly in short sessions without needing the user to sign a new UN.
However, this can also be a security risk in some contexts. 
This task is about allowing an optional extra parameter to be used in the cryptographic process of constructing a UN, which can be used to link the UN to a specific context, which can also locally be verified by a verifier.

weiwu-zhang:
Unclear. Do you mean the UN should contain the hash of the current website domain or URL?

jot2re:
This is more a "nice to have".
I talked with @oleggrib about this, and currently a UN can be reused during its entire validity. We discussed that this is generally not the functionality we actually want. We generally want it to just work a *single* time. However, this does not seem to currently be a big security issue.

oleggrib:
@jot2re , 
Current UN limited by:
- TTL
- domain 

in most cases UN will be used in useEthKey object inside website session. So we can replace domain with session nonce and use UN as is. If no session then we can use domain, not nonce.
But websites, who can allow access to something or sell something with discount use cookie-sessions to stay user cart, auth, etc syncronized., so we can get nonce and pass it to UN instead domain, because domain is just string.

oleggrib:
> But it is a bit more pretty and easier to read/understand using parametrisation.

If single param enough for use then we can rename __domain__ to __nonce__ and sometime pass session nonce, sometime domain

PR #284, opened 2022-07-12
feat: signature-based UN and possibility of adding context to UN

jot2re:
This PR adds another type of UN, where is signature-based and hence can be validated on-chain. 
The PR also adds the option of adding a context to a UN as well. 
This closes issue #275 and issue #259


jot2re:
Note that I have changed the size of the UN from 64 bits to 128 bits due to security issues. 
The code should have been changes such that old UNs will still be valid.

jot2re:
@micwallace can you have a look at the PR whenever you have time? In particular the JS code. Feel free to make improvements  to the JS since I am not experienced in that language

micwallace (approved):
@jot2re This looks good to me Tore, please feel free to merge and I'll get started with adding UN signature method to SafeConnect as part of the security review. We should also think about updating crypto-verify as well.

On customizing the signing message, I'll add this into SafeConnect for now, as a pathfinder for finding the right way to do it in this library. I forgot that it's actually in crypto-verify at the moment rather than this library. When your Java verification server is ready we will need to have a way to customize this at the application level. Maybe passing a simple template string as a parameter is the way to go.

PR #289, opened 2022-08-01
Feature: dynamic asn parser

micwallace:
Hey @weiwu-zhang 

Here is an examples of how the asn1-schema library can be utilized to serialize & parse ASN using a schema provided at runtime. 
The schema input is currently in JSON, but we can easily bridge the gap from .asd by writing an XML converter that can output this structure. Choice functionality will also need to be added to provide full support for ASN1.

### DIFF
--- a/src/main/javascript/crypto/package.json
+++ b/src/main/javascript/crypto/package.json
@@ -11,6 +11,7 @@
     "pack": "webpack",
     "clean": "rm -R dist",
     "watch": "webpack --watch --progress",
+    "serve": "webpack serve",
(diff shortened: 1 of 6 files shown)


jot2re:
@micwallace what is the status on this PR? It seems to becoming stale 😅

micwallace:
@jot2re We don't really have a use case at this stage but it will be needed if we want to integrate attestations into tokenscript for example.

jot2re:
@micwallace thanks for the update Miccy! I will keep the PR open then.

PR #297, opened 2022-09-19
Various fixes & enhancements for Devcon use cases

micwallace:
Hey @oleggrib 

Please review these changes when you get a chance:

- Fix: Issuer keys object being overwritten when importing into KeyPair objects. 
- Feature: return AttestedObject from Authenticator.validateUseTicket on successful validation. 
  This allow the validator to access attestation ticket data without subsequent decoding of the attestation. 
- Feature: allow validateUseTicket without userEthKey check by explicitly passing null for userEthKey argument.

### DIFF
--- a/src/main/javascript/crypto/package.json
+++ b/src/main/javascript/crypto/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tokenscript/attestation",
-  "version": "0.3.9-sc.6",
+  "version": "0.3.9-dc.1",
   "description": "A library for integrating cryptographic attestations into applications",
   "type": "commonjs",
   "main": "dist/index.js",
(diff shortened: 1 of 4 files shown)


micwallace:
As pointed out by @jot2re it seems I have missed some commits from this PR.

micwallace:
As per previous comment, the first two point in the description have been accidentally committed to the safe-connect branch and have already been merged into main.

oleggrib (approved):
> Feature: allow validateUseTicket without userEthKey check by explicitly passing null for userEthKey argument.
just note: function use default value null, so if skip agrument then it will as argument ===null

update is pretty simple , so all good

jot2re (approved):

PR #298, opened 2022-09-30
Candidate for 0.4

jot2re:
Contains multiple fixes and enhancements, like updates to the UNs and the SafeConnect code.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ configurations {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


jot2re:
There has been a lot of changes and fixes so I would like to make a new release of attestation.jar
... Although I would like to wait with pushing the release until after DevCon of course :)

jot2re:
Note that blockchain integration tests have been disabled to ensure that tests works, until the integration tests get updated to use current test-net

PR #299, opened 2022-10-08
Devcon

micwallace:
This adds support for validation against multiple public keys for a single conference ID


oleggrib (approved):
It works! thank you.

jot2re:
@micwallace it looks good, but would be nice with a negative test validating that keys from "wrong" conferences cannot be used to verify. I.e. that a key for Devcon 6 cannot be used to validate a signature on a ticket issued for Devcon 5.

jot2re:
Note that the reason the tests fail is not because of the change, but because the smart contract integration tests fails due to obsolescence of Rinkeby

oleggrib (approved):
looks good, thank you, @micwallace

PR #302, opened 2022-11-04
fix: setup things for new network

jot2re:
Still needs fix


jot2re:
Fixes #302

jot2re:
Handing this over to @JamesSmartCell

PR #306, opened 2022-11-17
fix: inverted if statement

micwallace:
Also: 

chore: update Goerli contract address.
fix: by default shift validFrom back to account for block.timestamp variance & delays.


jot2re (approved):
Looks good!

PR #310, opened 2023-05-22
fix: various fixes for EAS

micwallace:
- Include domain info in EasTicketAttestation.getDerEncoding. 
  This is required for EIP712 signature validation when UseTicket attestation is used on the blockchain. 
- Make userEthKey param optional in Authenticator.validateUseTicket() to be consistent with EASZKProof.create params. 
- Ensure new commitment secret is generated for each attestation. 
- Fallback to default key when conference ID field is not in use

### DIFF
--- a/src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
+++ b/src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
@@ -269,7 +269,7 @@ export class EasTicketAttestation extends AttestableObject implements Attestable
 		if (!uid)
 			uid = this.getEasUid();
 
-		const eas = new EAS(this.EASconfig.address, {signerOrProvider: this.signer});
+		const eas = new EAS(this.signedAttestation.domain.verifyingContract ?? this.EASconfig.address, {signerOrProvider: this.signer});
 
 		const revoked = await eas.getRevocationOffchain(this.signerAddress, uid);
 
@@ -328,7 +328,7 @@ export class EasTicketAttestation extends AttestableObject implements Attestable
 		this.loadEasAttestation(decoded.sig as SignedOffchainAttestation, keys, commitmentSecret)
 	}
 
-	getAsnEncoded(includeDomainInfo = false, compressed = false){
+	getAsnEncoded(includeDomainInfo, compressed = false){
 
 		const abiEncoded = defaultAbiCoder.encode(
 			this.signedAttestation.types.Attest.map((field) => field.type),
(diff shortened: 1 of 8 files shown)


src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
```diff
@@ -111,6 +111,9 @@ export class EasTicketAttestation extends AttestableObject implements Attestable
 		if (!this.signerAddress)
 			new Error("Failed to get signer address");
 
+		// New secret generated for each attestation
+		this.commitmentSecret = undefined;
```
oleggrib, line 115: I've added commit - remove extra check in line 125

src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
```diff
@@ -225,12 +228,15 @@ export class EasTicketAttestation extends AttestableObject implements Attestable
 		return this.decodedData;
 	}
 
-	getAttestationField(fieldName: string){
+	getAttestationField(fieldName: string, suppressException = false){
```
oleggrib, line 230: Hi, @micwallace , could you explain me reason why do we need to allow some empty fields?

src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
```diff
@@ -436,7 +442,7 @@ export class EasTicketAttestation extends AttestableObject implements Attestable
 
 	private processKeysParam(keys?: KeysArray){
 
-		let conferenceId = this.getAttestationField("devconId");
+		let conferenceId = this.getAttestationField("devconId", true);
```
oleggrib, line 444: @micwallace , same here, I cant get logic why its better to have fallback to key ""(empty string)? thanks

oleggrib:
Hi, @micwallace , I have added check to make sure we test **checkRevocation()** only if this Attestation is revocable. For now revocable is hardcoded, but it can be changed later.

src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
micwallace: I added the idea of having a default key for all eventIds, this was mainly for testing purposes.
```
base64senderPublicKeys: {
   "":"publickey pem..."
}
```

When a key for a specific isn't found, it will try to use the default key for validation. But this parameter is now redundant as I am using getAttestationData instead: 
src/main/javascript/crypto/src/eas/EasTicketAttestation.ts:445

Just pushed an update to remove it.

micwallace: @oleggrib I added the concept of a default key that can be used when a key for the specific devconId isn't set explicitly. Or for attestations that don't use eventId or devconId. This was initially adding for testing purposes.
```
"base64senderPublicKeys": {
   "": "PEM public key..."
}
```

I'm using getAttestationData() for this now so have removed the redundant suppressException parameter.

oleggrib (approved):
now it looks better, we can merge it

micwallace:
@oleggrib Can you please merge this? I can't bypass failed checks.

oleggrib:
merge to main

PR #311, opened 2023-06-21
feat: 🎸 added option to read OpenSSL key, EAS base64->base64url

oleggrib:
Changed EAS ticket encoding to base64url, because parsers incorrectly decode ticket in base64 format


micwallace (approved):
LGTM

PR #312, opened 2023-07-07
Feat: eas identifier attestation

micwallace:
Implements an EAS version of attestation.id identifier attestation, 
plus functionality to create and verify a ZKProof attestation from an EAS ticket and email attestation.


src/main/javascript/crypto/src/eas/EASIdentifierAttestation.ts
```diff
@@ -0,0 +1,54 @@
...
+		if (!options)
+			options = {};
+
+		// Identifier attestation is not revocable at this stage.
+		if (options.revocable === undefined)
+			options.revocable = false;
+
+		if (!options.validity){
+			const date = new Date();
+			date.setDate(date.getDate() + 30);
```
oleggrib, line 48: better add input variable for TTL of attestation  in seconds so user developer can easy change it.
@micwallace

src/main/javascript/crypto/src/eas/EasAttestedObject.ts
```diff
@@ -0,0 +1,112 @@
...
+import {EASIdentifierAttestation} from "./EASIdentifierAttestation";
+import {AttestationCrypto} from "../libs/AttestationCrypto";
+import {ProofOfExponentInterface} from "../libs/ProofOfExponentInterface";
+import {hexStringToUint8, logger, uint8tohex} from "../libs/utils";
+import {AsnParser, AsnProp, AsnPropTypes, AsnSerializer} from "@peculiar/asn1-schema";
+import {KeyPair, KeysArray} from "../libs/KeyPair";
+import {UsageProofOfExponent} from "../libs/UsageProofOfExponent";
+import {DEBUGLEVEL} from "../config";
+
+export class EasUseToken {
```
oleggrib, line 11: would be great to move this ASN class to the file in /asn1 folder to they them in one place

src/main/javascript/crypto/src/eas/EasAttestedObject.ts
```diff
@@ -0,0 +1,112 @@
...
+		if (!this.crypto.verifyEqualityProof(attCom, objCom, pok)) {
+			throw new Error("The redeem proof did not verify");
+		}
+		return pok;
+	}
+
+	getEncoded(){
+
+		const useToken = new EasUseToken();
+		useToken.ticketAttestation = new Uint8Array(this.ticketAttestation.getAsnEncoded());
```
oleggrib, line 63: @micwallace , no need to use ASN1 encoding in this case, better stay it solid and encode everything with defaultAbiCoder, like internal data encoded
in this case we dont need to use separate lib to decode ASN. it can save GAS in contract execution and save GAS when we deploy smaller contract.

src/main/javascript/crypto/src/eas/EASIdentifierAttestation.ts
```diff
@@ -0,0 +1,54 @@
...
+				{name: "identifierType", type: "string"},
+				{name: "commitment", type: "bytes", isCommitment: true},
+				{name: "ethereumAddress", type: "address"}
+			]
+		};
+
+		super(schema, signingConfig, undefined, issuerKey ? <KeysArray>{"": issuerKey} : undefined);
+	}
+
+	async createEasAttestation(data: EasIdentiferParameters|any, options?: EasTicketCreationOptions, commitmentType = 'mail'){
```
oleggrib, line 37: @micwallace commitment must be input value, it must be received from user side and commitmentSecret must stay in requestor side.

in current case some person, who has access to attestation.id privateKey can use it to issue attestation for ANY email. so its completely unsecure. Just check ASN version of idAttestation - it receive requestAttest and parse it for commitment and recovers userAddress

oleggrib (changes requested):

src/main/javascript/crypto/src/eas/EASIdentifierAttestation.ts
micwallace: @oleggrib validity from and to date. This can be easily changed by the developer by passing these options. By default, for identifier attestations the validity is 30 days as per the current ASN implementation.

src/main/javascript/crypto/src/eas/EasAttestedObject.ts
micwallace: I'd rather keep it here or as you said below, move to fully abi encoded structure.

src/main/javascript/crypto/src/eas/EasAttestedObject.ts
micwallace: Yep I agree. Keep in mind the contracts will require updating too. Can you help with that if I go ahead and make this change?

src/main/javascript/crypto/src/eas/EASIdentifierAttestation.ts
micwallace: @oleggrib Yeah but surely if the user has attestation.id private key we have bigger problems! Is there any value in keeping this client-side. What if the server just generates a random secret and creates commitment for any email address? Generating secret client side doesn't seem to offer any additional protection. Server can always issue attestations to any address it wants.

src/main/javascript/crypto/src/eas/EasAttestedObject.ts
micwallace: @oleggrib I've added an ABI encoded version. This is used by default but I've kept the ASN version around until the next major version.

src/main/javascript/crypto/src/eas/EASIdentifierAttestation.ts
oleggrib: updated code looks great.

only the question in commitment+secret+userWalletAddress.

current attestation.id version generates commitment+secret in userBrowser and signs request. it solves few problems:
1. no chance to make mistake in walletAddress, because backend recovers walletAddress from signature. backend issues attestation to the wallet, what user really controls. no chance to use wrong/fake address by mistake

2. id attestation secret stores in user browser only and dont walk through the network. it means nobody can grab network data and get idAttestation

what do you think, @micwallace ?

micwallace: Hey @oleggrib 

1. We still recover signature from Eip712AttestationRequest in order to set the correct eth address in the attestation, but this is done by common API instead of the library. You can review this code here: https://github.com/AlphaWallet/common-api/blob/85f9ec78a292c9f7cfc7547acd89f37c5509ff2a/backend/src/services/idAttestationsService.ts#L50C75-L50C75

2. We always use HTTPS so it's not possible to intercept the secret. In some ways it is better if there is no way for the client to specify the secret (and choose a weak one).

oleggrib (approved):
Thnak you, great, we can merge.

PR #316, opened 2023-09-06
Upgrade: eas v1 offchain support

micwallace:
NOTE: Merge https://github.com/TokenScript/attestation/pull/312 first!!!

- Update SDK and add support for v1 offchain attestations, with v0 backward compatibility
- Migrate URL encoding functions to use latest EAS SDK and remove AttestationURL.ts
- Return ethers tx object from revocation functions


src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
```diff
@@ -350,14 +368,14 @@ export class EasTicketAttestation extends AttestableObject implements Attestable
...
-		)
+		);
 
 		const attestation = defaultAbiCoder.encode(
 			this.signedAttestation.types.Attest.map((field) => field.type),
 			this.signedAttestation.types.Attest.map((field) => this.signedAttestation.message[field.name])
 		);
 
-		const signature = joinSignature(this.signedAttestation.signature);
+		const signature = joinSignature(this.signedAttestation.signature ?? this.signedAttestation);
```
oleggrib, line 378: Hi, @micwallace , could you explain this change? 
this.signedAttestation is SignedOffchainAttestation but joinSignature need SignatureLike param

src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
```diff
@@ -449,35 +467,45 @@ export class EasTicketAttestation extends AttestableObject implements Attestable
...
-			name: "EAS Attestation",
-			version: domainDecoded[0],
-			verifyingContract: domainDecoded[1],
-			chainId: domainDecoded[2]
-		}
-
 		// console.log("ABI Encoded bytes: ", "0x" + uint8tohex(new Uint8Array(asnEmbedded.easAttestation)));
 
+		const firstElem = defaultAbiCoder.decode(['uint16'], parts.attestation);
+		const version = firstElem[0] > 50 ? 0 : 1;
```
oleggrib, line 473: @micwallace , why do you use 50 as limit? in current implementation we allow only version 1,
so to check more bit and avoid collision when schema last bytes are 0x0001 , better use
```
const firstElem = defaultAbiCoder.decode(['uint256'], parts.attestation);
const version = firstElem[0] > 50n ? 0 : 1;
```
it will cover more bytes

src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
```diff
@@ -541,9 +569,9 @@ export class EasTicketAttestation extends AttestableObject implements Attestable
 			chainId: this.signedAttestation.domain.chainId
 		};
 
-		const offchain = new Offchain(config);
+		const offchain = new Offchain(config, this.signedAttestation.message.version ?? 0);
 
-		const hash = ethers.utils._TypedDataEncoder.hash(offchain.getDomainTypedData(), {Attest: ATTESTATION_TYPE}, this.signedAttestation.message);
+		const hash = ethers.utils._TypedDataEncoder.hash(offchain.getDomainTypedData(), {Attest: OFFCHAIN_ATTESTATION_TYPES[this.signedAttestation.message.version].types}, this.signedAttestation.message);
```
oleggrib, line 574: @micwallace , to make code more beautiful we can read string "Attest" from OFFCHAIN_ATTESTATION_TYPES[this.signedAttestation.message.version].primaryType

although this value doesnt matter here... we can use any name here and it will work correct

src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
micwallace: In a previous version of the signed attestation JSON, the r,s,v parameters were in their own signature property (making it compatible with SignatureLike type). Later they were moved into the top level object. This change is just for backward compatibility.

src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
micwallace: Honesty I didn't know how to read the length of the ABI element, so I just went with a safe default lol.
I've applied this change it's working well.

src/main/javascript/crypto/src/eas/EasTicketAttestation.ts
micwallace: Actually it doesn't work, seems there is a bug there in the EAS SDK. Even though v0 is specified with primaryType=Attestation, the attestations are actually issued with types={Attest: [...]} just like v1.

Relevant SDK code:
```
exports.OFFCHAIN_ATTESTATION_TYPES = {
    0: {
        domainName: 'EAS Attestation',
        primaryType: 'Attestation',
        types: [
            { name: 'schema', type: 'bytes32' },
            { name: 'recipient', type: 'address' },
            { name: 'time', type: 'uint64' },
            { name: 'expirationTime', type: 'uint64' },
            { name: 'revocable', type: 'bool' },
            { name: 'refUID', type: 'bytes32' },
            { name: 'data', type: 'bytes' }
        ]
    },
    1: {
        domainName: 'EAS Attestation',
        primaryType: 'Attest',
        types: [
            { name: 'version', type: 'uint16' },
            { name: 'schema', type: 'bytes32' },
            { name: 'recipient', type: 'address' },
            { name: 'time', type: 'uint64' },
            { name: 'expirationTime', type: 'uint64' },
            { name: 'revocable', type: 'bool' },
            { name: 'refUID', type: 'bytes32' },
            { name: 'data', type: 'bytes' }
        ]
    }
};
```

micwallace: The issue seems to be here:
```
const signedRequest = await this.signTypedDataRequest(params, {
            domain: this.getDomainTypedData(),
            primaryType: this.type.primaryType,
            message: params,
            types: {
                Attest: this.type.types
            }
        }, signer);
```

But this bug is even in older versions of EAS SDK, so it's nothing to worry about. 
You can see it for newly issued attestations here using an older library version: https://create.openloyalty.com/

oleggrib (approved):
Thanks @micwallace  , I am happy with other code.

